PR #4, opened 2019-11-20
Allow adding multiple criterias.

fiseni:
Allow adding multiple criterias in the constructor body, by calling AddCriteria(criteria) method.
This would offer the following advantages
- Adding criteria (where clause) based on some condition.
- Adding multiple conditions in more clean fashion, instead of using "&" in a long expression as a constructor parameter.

In addition, having parameterless constructor, would offer creating empty specifications. This might be quite useful feature. Whenever we need list of all records, we can start with empty specification. If afterward, we require some global filter for particular entity, we can add the criteria in the empty specification.


ardalis:
Thanks!

PR #9, opened 2019-12-14
Feature/chainable includes

mrukas:
I added the possibility to use .Include followed by .ThenInclude. This should make it easier to define what properties should be loaded.
Chainable Includes add their paths to the existing IncludeStrings, because i think the most generic approach to define proprties to load are strings.


ardalis:
Thanks!

PR #13, opened 2020-01-07
Consolidating the selector functionality.

fiseni:
Hi @ardalis 

I added the missing parts and the integration tests for the selector.

Let me describe shortly the functionality:

- The intention was not only adding the selector functionality in the specification, but also keeping the evaluation of the same within the specification (not delegating it out to the EFRepository).
- Select always returns new type. That's why need new inherited infrastructure containing the output type.
- Adding new List method in EFRepository would just confuse the users. Therefore, the method is overloaded, and based on the fact if specification contains a selector or not, the appropriate method will be utilized.

### DIFF
--- a/src/Ardalis.Specification/BaseSpecification.cs
+++ b/src/Ardalis.Specification/BaseSpecification.cs
@@ -7,8 +7,6 @@ using System.Linq.Expressions;
 
 namespace Ardalis.Specification
 {
-    // Keeping both classes here for clarity. Will fix it afterward
-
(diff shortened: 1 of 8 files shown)


fiseni:
I'm not quite sure why the build failed in azure pipeline. It failed to compose the docker image (image operating system "linux" cannot be used on this platform).
Azure just recently this week changed the build pipeline, so the configuration might need to be revisited.

ardalis:
Thanks!

PR #15, opened 2020-02-06
merge branch master of MoienTajik/Specification with ardalis/Specification

MoienTajik:
- Fixed some typos and applied some rename refactorings
- Removed unused using statements from project
- Updated test projects to .NET Core 3.1
- Upgraded test projects packages
- Migrated IntegrationTest's EF underlying provider from SqlServer to in-memory database
- And some little enhancements ...

### DIFF
--- a/src/Ardalis.Specification/BaseSpecification.cs
+++ b/src/Ardalis.Specification/BaseSpecification.cs
@@ -2,7 +2,6 @@
 using Ardalis.Specification.QueryExtensions.Include;
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Linq.Expressions;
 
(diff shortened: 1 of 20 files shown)


ardalis:
Thanks for the PR @MoienTajik. I prefer to keep `protected` fields prefixed with `_` even though I know that doesn't correspond with  official guidance, which uses the same format for `public` and `protected`. `protected` is much closer to `private` than it is to `public`. Constants also should be ALL_CAPS. This got old so there are conflicts (my fault -sorry). If you want to resolve them, pull in the latest master, and stick to the conventions I'm using, I'd take the PR. If not I understand. Thanks!

PR #16, opened 2020-06-03
Adding ThenBy functionality to chain with OrderBy

efleming18:


ardalis:
Can you look at why some of the integration tests fail in the docker container?

image

I don't think it's your PR but I've been trying to get that fixed for a little while now.

PR #21, opened 2020-06-09
Enable nullable reference types

deepumi:
Enable C# 8.0 Nullable Reference Types for the following projects
1)  Ardalis.Specification
2)  Ardalis.Specification.IntegrationTests
3)  Ardalis.Specification.UnitTests

Locally I tested both unit test and integration test and looks like all test passed successfully!

image

@ardalis  Please take a look.

### DIFF
--- a/src/Ardalis.Specification/SpecificationEvaluator.cs
+++ b/src/Ardalis.Specification/SpecificationEvaluator.cs
@@ -1,12 +1,11 @@
-﻿using Microsoft.EntityFrameworkCore;
-using System.Linq;
+﻿using System.Linq;
 using System.Threading.Tasks;
 
 namespace Ardalis.Specification
 {
     public class SpecificationEvaluator<T> where T : class
     {
-        public static async Task<IQueryable<T>> GetQuery(IQueryable<T> inputQuery, ISpecification<T> specification)
+        public static Task<IQueryable<T>> GetQuery(IQueryable<T> inputQuery, ISpecification<T> specification)
         {
             var query = inputQuery;
 
@@ -47,7 +46,7 @@ namespace Ardalis.Specification
                 query = query.Skip(specification.Skip)
                              .Take(specification.Take);
             }
-            return query;
+            return Task.FromResult(query);
         }
     }
 }
--- a/src/Ardalis.Specification/SpecificationEvaluator.cs
+++ b/src/Ardalis.Specification/SpecificationEvaluator.cs
@@ -5,7 +5,7 @@ namespace Ardalis.Specification
 {
     public class SpecificationEvaluator<T> where T : class
     {
-        public static Task<IQueryable<T>> GetQuery(IQueryable<T> inputQuery, ISpecification<T> specification)
+        public static IQueryable<T> GetQuery(IQueryable<T> inputQuery, ISpecification<T> specification)
         {
             var query = inputQuery;
 
@@ -46,7 +46,8 @@ namespace Ardalis.Specification
                 query = query.Skip(specification.Skip)
                              .Take(specification.Take);
             }
-            return Task.FromResult(query);
+
+            return query;
         }
     }
 }
--- a/src/Ardalis.Specification/Ardalis.Specification.csproj
+++ b/src/Ardalis.Specification/Ardalis.Specification.csproj
@@ -19,7 +19,8 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Ardalis.GuardClauses" Version="1.5.0" />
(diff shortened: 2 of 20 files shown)


src/Ardalis.Specification/SpecificationEvaluator.cs
```diff
@@ -1,12 +1,11 @@
...
-using System.Linq;
+﻿using System.Linq;
 using System.Threading.Tasks;
 
 namespace Ardalis.Specification
 {
     public class SpecificationEvaluator<T> where T : class
     {
-        public static async Task<IQueryable<T>> GetQuery(IQueryable<T> inputQuery, ISpecification<T> specification)
+        public static Task<IQueryable<T>> GetQuery(IQueryable<T> inputQuery, ISpecification<T> specification)
```
ardalis, line 8: Why did you remove async here?

src/Ardalis.Specification/SpecificationEvaluator.cs
```diff
@@ -47,7 +46,7 @@ public static async Task<IQueryable<T>> GetQuery(IQueryable<T> inputQuery, ISpec
                 query = query.Skip(specification.Skip)
                              .Take(specification.Take);
             }
-            return query;
+            return Task.FromResult(query);
```
ardalis, line 49: Did you modify this due to a compiler warning, or ?

src/Ardalis.Specification/SpecificationEvaluator.cs
deepumi: @ardalis 
Good question, the method does not have any asynchronous operation like network calls and hence not required to add a 'async', However, if you want to consume on the other side we can still call the method like 'await GetQuery()' as it support Task<T>. Visual studio was showing warning that's why I have to make this change..

src/Ardalis.Specification/SpecificationEvaluator.cs
deepumi: @ardalis Yes, that's correct. Since the method does not do any actual async operation, call and be still synchronous operation.

ardalis: True. In your opinion is there value in returning a task at all, instead of just returning `IQueryable<T>` ?

deepumi: Well, bases on the scenario below, yes its recommended. 

In REST API call, has some async calls to get a network or database request and this case all other method making to async / await is beneficial since it won't make the incoming thread to wait to process a synchronous call.  And making the whole request pipeline is async/await is recommended.  (https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md)

If we don't require any asyc/await call for the method we can simply return IQueryable<T>. Since this library using some places and I don't recommend to make a breaking change to async to sync.

fiseni: This method originally was synchronous. The method should do only in-memory operations, it just evaluates different expressions and translates into iqueryable. Any async calls to external libraries within the evaluator might affect overall performance (specs should be created on the fly). Externall calls should be responsibility of the repository.

deepumi: That's correct, adding a "async" will generate extra code on the fly. Please check both links with 'async' and with 'no async'

With Async = https://sharplab.io/#v2:D4AQTAjAsAUCAMACEECsBuWsQGZlkQGFEBvWRC5PEAFkQFkAKASlPMoF92LvEAHAE4BLAG4BDAC4BTZBABsyABzIFACSkAbDQHsA6toEaAJgEEAzgE8AdgGMA+gFEAHhIFiWZGJW+JeFEACcKgB0AGIC2gC2AEpSZgCuGhKMAETqWtqIAO4GxinM6JS8XDAcQA==

With out Async = https://sharplab.io/#v2:D4AQTAjAsAUCAMACEECsBuW55bAZkQG9ZFTkwwiSyaUBOACgCJAeDcBB9pgSkxhoF9qpQYgAOAJwCWANwCGAFwCmyCADZkagBIKANtoD2AdT1jtAEwCCAZwCeAOwDGDTsV40yw0iADs6gHQAxMT0AWwAlBUsAV205Zi1dPUQAd2MzLnR3V1IBGD4gA==

My proposal was to get rid of async usage, so we don't want to create any unnecessary allocations..  And I really don't see much values to make this method as async.. since its consuming other projects this would be a breaking change as I mentioned above.

ardalis: Right I'm thinking it should just be synchronous and return IQueryable which eliminates all of the async/Task crud from it (and as noted it doesn't actually do any async work).

deepumi: Awesome... If you want I can take care of it as a new PR once this gets approved or merged!

deepumi: Changing Async to Sync will eventually affect following files.. 

image

ardalis: If you don't mind, go ahead and make the change to sync. We don't need any of that to be async at this point - it was changed while trying to get GroupBy working with EF Core 3.x.

deepumi: Sure no problem... I will commit the changes now

deepumi: @ardalis, I just committed my changes. Please review https://github.com/ardalis/Specification/pull/21/commits/9bdca6be7a61da1df540cad521a5cf281dfa78db

PR #46, opened 2020-08-28
Patch 1

ShadyNagy:
I think we need to release a new version for this changes #44


fiseni:
Thanks @ShadyNagy 
I intentionally didn't update the version numbers. Publishing new versions should be @ardalis responsibility, so this PR will wait for his approval.

Issue #50, opened 2020-10-08
Is there any way I can OrderBy a SmartEnum name?


RandomUser99:
Hi,

So I'm using both this library in addition to Steve's SmartEnum. I have a SmartEnum Status class in my .Net CORE 3.1 project:

 ```
public class BookingStatus : SmartEnum<BookingStatus>
    {
        public string ShortName { get; }
        public bool IsConfirmed { get; }
        public StatusUpdateType UpdateUpdateType { get; }

        public static readonly BookingStatus Submitted = new BookingStatus("Booking request submitted", "Submitted", 0, false);
        public static readonly BookingStatus Confirmed = new BookingStatus("Booking confirmed", "Confirmed", 1, true);
        public static readonly BookingStatus Cancelled = new BookingStatus("Booking cancelled", "Cancelled", 2, false);
        public static readonly BookingStatus ConfirmedDateChanged = new BookingStatus("Confirmed date amended", "Confirmed date amended", 3, true, StatusUpdateType.LogOnly);

        private BookingStatus(string name, string shortName, int value, bool isConfirmed) : base(name, value)
        {
            ShortName = shortName;
            IsConfirmed = isConfirmed;
            UpdateUpdateType = StatusUpdateType.LogAndUpdate;
        }
    }
```

I also have a Specification class which I want to OrderBy the BookingStatus.ShortName property. EF doesn't like this in 3.1 due to the error "Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly".

I could do this client side using LINQ, but I am using the Skip() and Take() feature of this library for pagination, so I cannot order client side as the results set is incorrect

Is there any way around this or force my query to use client side evaluation or another solution possibly?

Thanks.

**Comment by fiseni** [comment]
Hi @RandomUser99,

I haven't used SmartEnum extensively and I don't know if it has support for EF persistence, @ardalis will answer you on that.
As for the specifications, whatever you define within a spec, everything will be sent to the ORM, and the translation is done on the ORM. So, you have to be careful while building the specs, and be aware what can be translated to an actual query and what not. The last version of EF Core 5, is quite powerful and will translate quite complex stuff. But, still there are limitations.

Lately, based on my own experience, I've been thinking to incorporate the capability for InMemory operation in specifications. As a matter of fact, the specifications will just hold the information what to be executed after data is retrieved from persistence.
The usage might something like this
```
public class CompanySpec : Specification<Company>
{
    public CompanySpec(int countryId)
    {
[...]

PR #56, opened 2020-10-27
Implementation of InMemory feature for post manipulation of retrieved data.

fiseni:
Now that other PRs are merged, this is one more feature that is mentioned in #50
It's quite straightforward, the specification holds `Func<List<T>, List<T>>` predicate, and this is callled after the data is retrieved from persistence.
The usage is as following:

```
public class CompanySpec : Specification<Company>
{
    public CompanySpec(int countryId)
    {
        Query.Where(x => x.CountryId == countryId)
             .Include(x => x.Stores);

        Query.InMemory(x =>
        {
            // Here you are not constrained to the specification builder.
            // You can use everything that .NET offers in order to manipulate List
        });
    }
}
```

I must admit, at first glance this seemed quite appealing feature, but I'm having second thoughts on that :)
Not a bad thing to have it, but if thought carefully very rarerly will be used. If you're in a position that you need additional filtering after the DB query is completed, then it means you have quite inefficient query.
One use case, since `IncludeWhere` is not supported until EF Core 5, then you might need additionally filtering based on some complex conditions on the navigations? Maybe?

I don't know, the community might find some other use cases.


fiseni:
As soon I wrote the previous comment, an apple fall on my head :)

If you love domain events and use them often in your code, then most probably you always have trouble to decide where exactly to subscribe to them? Imagine the aggregate root have to subscribe to its child's event. Now you can just iterate through the collection and call subscribe for each retrieved aggregate root. As a final result, you get list of aggregates from repository who are already completely set.

Note: This is actually more valid if you're using .NET events. If you have your own event aggregator or similar infrastructure, then there are other ways to this problem.

fiseni:
If we're ok with the naming, this can be merged too (resolved the conflicts now).

wrakocy:
Hi @fiseni . Wondering if the InMemory feature was ever released? I'm pulling in version 5.1, but not seeing it. Thanks!

fiseni:
Yes, it's there, but renamed to `PostProcessingAction`.
We did that to avoid confusion with in-memory evaluators in version 5.

Sorry, we have not been very careful in documenting all changes. We're working on the documentation.

wrakocy:
Ah! OK, perfect. That'll do just fine ;) Thanks for the assist!

PR #62, opened 2020-11-14
Added project Ardalis.Specification.EntityFramework to support EF6

ngruson:
A solution I'm working on is still using some WCF services on .NET Framework 4.7.2 with Entity Framework 6.
A while ago, I made a repository that is depending on EF6 instead of EF Core.
Here's the PR.


fiseni:
Hi @ngruson,

Great, @ardalis will comment and review this. 
I'll just give a few notes to consider:
- Should TFM be "v4.7.2"? It might be better if we create ".netstandard2.0" project. EntityFramework6 will work just fine with it, and we won't restrict it to only 4.7.2 and higher.
- A naming convention for the plugin/addon packages.
I'm not sure, but what if we add "6" to the end? Indeed, "EntityFramework" is the official name, but while installing you are aware of version "v6". In our case, it will be something like "Ardalis.Specification.EntityFramework v1.0". Is this clear enough?

Now that the `EF Core5` is out there, we might have the following packages:
- Ardalis.Specification.EntityFramework6
- Ardalis.Specification.EntityFrameworkCore (Probably we did a mistake here, should have named it "Core3" instead)
- Ardalis.Specification.EntityFrameworkCore5

ngruson:
Hi @fiseni, 

Thanks for the comments.
- Initially I tried .NET Standard as TFM. EntityFramework 6.4 is depending on .NET Standard 2.1, not on .2.0.
.NET Framework is not compatible with .NET Standard 2.1 (https://docs.microsoft.com/en-us/dotnet/standard/net-standard). Therefore I had to fall back to .NET Framework 4.7.2. Now I've upgraded to 4.8.
- I agree with the naming convention of EntityFramework6.

I noticed also that I was behind compared to Ardalis.Specification.EF which has had some file additions. Might as well add ParameterReplacerVisitor.cs and SearchExtension.cs while we're at it.

ardalis:
@ngruson If you don't mind, can you pull down latest and get the solution file conflicts resolved?

ardalis (commented):
This looks good once the solution file is fixed up.

ngruson:
@ardalis Did some merging and now it should be ok.

fiseni:
Thanks @ngruson,

Some of the stuff should be copied once again from `EFCore` project to reflect the new changes, such as the `SpecificationEvaluator`. No longer the base evaluator should be used (in order to preserve the correct ordering). Also `SearchEvaluator` is not added.

Anyhow, we can merge it, and then fix these details.

Thank you.

ngruson:
I guess you're referring to SearchExtension.cs?
The Search extension method contains some EF Core dependent code.
Not sure what the EF6 equivalent would look like.

fiseni:
Yes, sorry I meant `SearchExtensions`.

It's OK, we'll fix it afterward.

Thank you.

ardalis:
@fiseni any issue with me merging this one? I'm not using EF6 which is the only reason I've been slow to pull these changes in.

fiseni:
It can be merged, but we have to work on it afterward.
Maybe, initially, we can clarify how we want to proceed with these plugin packages. See issue #62 .
Once we decide, then it's easy to move forward.

fiseni:
Hey @ngruson,
We have a new version published. Would you still like to add support for EF6. We'll have to add the existing PR accordingly.
If you're not available it's totally fine, we'll try to pick it up. 

Thank you for your support :)

ngruson:
Would be nice to include EF6 support, I'm working on it.
Also adding unit tests similar to the EF Core unit tests you already have.

ngruson:
EF6 support has now been added.
@fiseni : Can you have a look?

The tests for EF6 and EF Core are passing, but the build is failing since it does not support .NET Framework.
How would you go about this? We can put some filters in the yaml to exclude the .NET Framework assemblies from the normal build and add another step to build the .NET Framework bits with MSBuild.

Furthermore:
1) I also had to change OrderEvaluator to accommodate EF6.
2) In the tests for paginated specs, EF6 requires an OrderBy call before Skip and Take are being called. If you don't want this for the EF Core spec tests, we can add separate specs for EF6 in the unit test project.

fiseni (changes requested):
Hey @ngruson,

Great work, thank you!

Please, if possible do the following changes.
- Lets not change the `OrderEvaluator` in the base package. You can define a new `IEvaluator` implementation in the EF6 package for Order, and then use that one in the `SpecificationEvaluator` in EF6 package.
- Yes, please let's exclude it in yaml. We'll define a new build task for it later on.

For the `OrderBy` call before `Take` and `Skip`, no problem at all. Anyway `EF Core` adds ordering by id, if it's not explicitly defined.

Awesome job, btw.

fiseni (approved):

fiseni:
Thank @ngruson. I resolved the conflicts and it can be merged. 
@ardalis do you have any comments here?

ngruson:
In Ardalis.Specification.UnitTests.csproj, the target framework net472 should be removed, I think?
Running the integration tests in Docker now fails.

fiseni:
Yea, unit tests are not related to EF at all, so `net5.0` should be OK. I wasn't sure why you have added it, was planning to check it after the merge :)
Also, you didn't have to implement `IInMemoryEvaluator` either, `IEvaluate` is enough. The in-memory evaluator would work from the base package.

fiseni:
@ardalis, if you're OK here, let's merge it.
Once it is merged, we'll fix a few minor stuff.

ardalis:
Thanks @ngruson !

fiseni:
Fixed the issues and the build is working now (for `net5.0` runtime only, excluding the EF6 projects).
@ngruson do you know how to publish Nuget package from this? Do the package information tags (PackageId, Title, etc) work on the old .csproj files? If you have something ready that would be great, otherwise, we'll have to research and test the process.

ardalis:
As I recall we just need a .nuspec file for .NET Framework libraries.

https://docs.microsoft.com/en-us/nuget/reference/nuspec

PR #74, opened 2021-01-24
Refactoring tests.

fiseni:
The PR contains refactoring and a complete overhaul of the testing infrastructure and the tests (bunch of files are removed, and even more added). The coverage is around 90 %.

The idea is the following:
- Test the actual behavior that the package is offering. For an instance, test if the specification is populated correctly while building it. Testing if the expressions are doing the correct thing is a trivial test (since we're not modifying them).
- Create only unit tests for the base package. Test each builder method. On top of that, few specification tests are provided just for illustration purposes.
- Remove integrations tests for the base package.
- Create integration tests for the plugin/addon packages. Test the repository methods, thus we'll cover the repository and the specification evaluation tests.
- Standardize the naming conventions.
Folder Name:		        [Feature or Namespace]Tests
File Name: 			[Class Name]_[Method Name]
Test Method Name: 	[Retur/Action Information]_[Given Some Input/Condition]

Also, the unit test project serves as a common project for tests, it contains the fixture (entities, spec, etc). The integration test project (and other projects in the future for different providers) will reference and use the same fixture.

Note: It's already squashed and merged on my end, so the PR contains only 1 commit.


ardalis (approved):

ardalis:
Wow, awesome! Thanks!

PR #89, opened 2021-02-26
Refactoring for version5

fiseni:
This PR covers all the issues/features that we planned for version5. Also, the plugin package is updated to `.NET Standard 2.1` and `EF Core 5`. It contains many commits but tried to be descriptive in the messages (not fully independent though, features are inter-related and some of the commits fix previous issues, etc).

Note: Added non-generic repository and interface too. For now, made a full copy and adapt it. In order to decorate, we should mark the non-generic one as non-abstract, and not quite sure about that.
The main difference in the usage would be `GetById` methods. Since the `T` can't be inferred, should be written explicitly while calling the method.
```c#
var result = await repository.GetByIdAsync<Store>(1);
```


ArdalisSpecification/src/Ardalis.Specification/IRepositoryBase.cs
```diff
@@ -10,7 +10,7 @@ namespace Ardalis.Specification
     /// </para>
     /// </summary>
     /// <typeparam name="T">The type of entity being operated on by this repository.</typeparam>
```
ardalis, line 12: Remove `<typeparam>`

ArdalisSpecification/src/Ardalis.Specification/IRepositoryBaseOfT.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// Finds an entity with the given primary key value.
+        /// </summary>
+        /// <param name="id">The value of the primary key for the entity to be found.</param>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The task result contains the <typeparamref name="T" />, or <see langword="null"/>.
+        /// </returns>
+        Task<T?> GetByIdAsync(int id);
```
ardalis, line 56: If we're using `TId` should we still have an `int` overload?
If we're going to have an `int` overload should we have a `Guid` and/or `string` overload too?

ardalis (changes requested):
Looks great - just a couple of suggestions

ArdalisSpecification/src/Ardalis.Specification/IRepositoryBaseOfT.cs
fiseni: I'm not sure even if we need the int overload. the generic version will pick up the type, and actually, it will be inferred, not that you will need to specify `<int>` part. So, probably we don't need any of the specific overloads.

ArdalisSpecification/src/Ardalis.Specification/IRepositoryBase.cs
fiseni: I didn't pay attention to the doc comments. We'll have to cross check everything in that regard. I'll remove this one now.

ArdalisSpecification/src/Ardalis.Specification/IRepositoryBaseOfT.cs
ardalis: Shorter it better, then.

fiseni: Should I remove it completely?

ardalis:
I think we should close this. My plan is to merge from v5 to master, once we're sure it's good to go.

fiseni:
Closing as a duplicate. There is new PR to branch `v5`

PR #90, opened 2021-02-26
Refactoring for version 5 (merging to "v5" branch).

fiseni:
Here is the same PR for the "v5" branch.
Not sure, but merging directly to master might be simpler at this point. It will create confusion for folks working on master (e.g. XML documentation).


ardalis:
Keeping it in a non-master branch will let me get a local copy of it easier. Maybe after this one goes out we can think about a dev branch that produces pre-release builds to nuget.org. For now I need to play with it a bit locally for my own projects

lankymart:
So if I update to v5, it's going to break any current `SpecificationEvaluator<T>`. Until we can update our repository I'm stuck with v4 and can't utilise filterable `Include()`.

lankymart:
We have a Generic Repo `EFRepository<T>` that uses;

```c#
private IQueryable<T> ApplySpecification(ISpecification<T> spec)
{
    var evaluator = new SpecificationEvaluator<T>();
    return evaluator.GetQuery(_dbContext.Set<T>().AsQueryable(), spec);
}
```
If we update the package to v5 it breaks because `SpecificationEvaluator` no longer accepts a generic.

fiseni:
If this is the only issue, then it can be easily fixed. You can do the following change. We refactored the evaluator to non-generic (instead it's generic on method level now) since anyhow it doesn't have an internal state and there is no need to create allocations. We tried to minimize the memory footprint here.

```c#
public IQueryable<T> ApplySpecification(ISpecification<T> specification)
{
  // You can also create an instance `new SpecificationEvaluator()`, but you can use the default singleton instance.

   return SpecificationEvaluator.Default.GetQuery(dbContext.Set<T>().AsQueryable(), specification);
}
```

But, please take a look at some new additional concepts provided in our base [repository](https://github.com/ardalis/Specification/blob/main/ArdalisSpecificationEF/src/Ardalis.Specification.EF/RepositoryBaseOfT.cs). If you like it, you can also use the shipped repository. All you need will be just to inherit from it.


```
public interface IRepository<T> : IRepositoryBase<T> where T : class
{
}

public class MyRepository<T> : RepositoryBase<T>, IRepository<T> where T : class
{
	private readonly MyDbContext dbContext;

	public MyRepository(MyDbContext dbContext) : base(dbContext)
	{
		this.dbContext = dbContext;
	}

	// Not required to implement anything. Add additional functionalities if required.
}
```

Btw, all repository methods are defined as virtual, so you can override whatever you don't like.

Issue #73, opened 2021-01-24
Improve AsNoTracking integration tests after EFCore5 upgrade


Clivar:
Integration tests for EFCore have all entities tracked by default as they are seeded, To make tracking tests work, you are required to manually detach the tested entity first. This is not a perfect solution and may produce side effects.

EFCore5 supplies `ChangeTracker.Clear` for cases like this which is much faster and side-effect free (for example while working with multiple entities or Include's)

Involved PR #72 
Documentation about [ChangeTracker.Clear Method](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.changetracking.changetracker.clear?view=efcore-5.0)

ardalis:
Thanks, we'll look at this once we have a package depending on EFCore 5.

fiseni:
Hey @Clivar,

Do you want to update your test now that we upgraded to `EF Core 5`? Your test is [here](https://github.com/ardalis/Specification/blob/3eeeeb66f070643d64570539e7681af15968dd20/ArdalisSpecificationEF/tests/Ardalis.Specification.EF.IntegrationTests/Repository_ListAsync.cs#L179). Please work with `v5` branch. Pull that branch and PR to that branch.

It doesn't hurt to have this kind of integration test, but I don't think it's crucial from the package's perspective. For us, it's important to test that if this feature is used then the flag is saved in the specification, and the evaluator passes the operation to the provider. That's all we do. The actual implementation is done on the provider's side (EF Core), and it's not our concern. On the other side, user's in their code-base might have full integration tests, and test for their scenarios.

fiseni:
If you'll be working on this, please add these unit tests. It's a simple one. Sounds OK?

```c#
public class SpecificationBuilderExtensions_AsNoTracking
{
    [Fact]
    public void DoesNothing_GivenSpecWithoutAsNoTracking()
    {
        var spec = new StoreEmptySpec();

        spec.AsNoTracking.Should().Be(false);
    }

    [Fact]
    public void FlagsAsNoTracking_GivenSpecWithAsNoTracking()
    {
        var spec = new CompanyByIdAsUntrackedSpec(1);

        spec.AsNoTracking.Should().Be(true);
    }
}
```

Clivar:
>  For us, it's important to test that if this feature is used then the flag is saved in the specification, and the evaluator passes the operation to the provider. That's all we do
[...]

PR #95, opened 2021-03-02
Migrate AsNoTracking tests to efcore 5.0

Clivar:
See: https://github.com/ardalis/Specification/issues/73#issuecomment-789280108


ArdalisSpecificationEF/tests/Ardalis.Specification.EF.IntegrationTests/RepositoryOfT_GetBySpec.cs
```diff
@@ -77,5 +77,17 @@ public async Task ReturnsCompanyWithStoresAndProducts_GivenCompanyByIdIncludeSto
             result.Stores.Count.Should().BeGreaterThan(49);
             result.Stores.Select(x => x.Products).Count().Should().BeGreaterThan(1);
         }
+
+        [Fact]
+        public async Task ReturnsUntrackedCompany_GivenCompanyByIdAsUntrackedSpec()
+        {
+            dbContext.ChangeTracker.Clear();
```
fiseni, line 84: We're using a shared database for the tests, all tests working with the same dbContext. In the case of some other test relying somehow on the tracker, clearing the tracker might lead to inconsistent results.

For now, we're good, but let's keep that in mind.

fiseni (approved):
We're good here (left a small note about clearing the tracker).

@ardalis, if you don't have any staged/working files locally, we can merge this from @Clivar .

PR #98, opened 2021-03-08
create specifications docs page

ilyanaDev:
"How to Create Specifications" docs page under "Usage"


ardalis (approved):

ardalis:
In another PR perhaps add to end of this and point out that in this case the Specification should probably also implement the https://github.com/ardalis/Specification/blob/master/ArdalisSpecification/src/Ardalis.Specification/ISingleResultSpecification.cs interface, since any `ById` specification should typically only return at most one result.

PR #101, opened 2021-03-10
Where page under base features

ilyanaDev:


docs/features/where.md
```diff
@@ -7,4 +7,10 @@ parent: Base Features
 grand_parent: Features
 ---
 
-# Where
```
ardalis, line 10: I think you need to keep the top level heading or it won't show anywhere on the page what this page's topic is.

ardalis (changes requested):
add heading

docs/features/where.md
```diff
@@ -7,4 +7,10 @@ parent: Base Features
 grand_parent: Features
 ---
```
fiseni, line 9: Hey @ilyanaDev 

You're doing a great job here, thanks for the help.
We might add the following sentence at the beginning, and then you can go and describe it in more detail. Folks using Linq are already quite familiar with this operation, so it will be clear for them.

> The `Where` feature defined in the specification behaves the same as `Where` in `Linq`, and it accepts `Expression<Func<TSource, bool>>` expression as a parameter.

We can start in the same manner for the other features too.

fiseni (approved):

docs/features/where.md
ilyanaDev: ah, yeah, you're right! Sorry, VS Code yelled at me for having a top-level heading, so I got rid of it. I'll fix that.

docs/features/where.md
ilyanaDev: sounds good

ardalis (approved):

PR #102, opened 2021-03-10
faq moved to faq in docs under getting started

ilyanaDev:


ardalis (approved):
This is good for now - we will update the FAQ once it is in place in the docs.

Issue #108, opened 2021-03-15
Add `CancellationToken` for the async repository methods.


fiseni:

fiseni:
Merged.

PR #109, opened 2021-03-16
Added cancellationToken for repository's async methods.

fiseni:
- Added cancellation token for the async methods
- Added constraint for `GetById` method, `where TId : notnull`


fiseni:
fixes #108

PR #110, opened 2021-03-18
Added Evaluate function to check if an entity fulfills to the specifi…

JelmarVA:
I'll start with an example why you even want to Evaluate against one entity.

Let's say I have a VerificationCodeService with SomeMethod:

``` c#
public async Task SomeMethod()
{
    var codes = await _verificationCodeRepository.ListAsync(new ValidVerificationCodeSpec());

    foreach (var code in codes)
    {
        if (somethingWillDestroyTheCode)
        {
            code.Destroy();
        }
        if (!code.IsValid())
        {
            //Do stuff for invalid case
        }
    }
}
```

I get all the verificationcodes that are valid using the ValidVerificationCodeSpec:
``` c#
public class ValidVerificationCodeSpec: Specification<VerificationCode>
{
    public ValidVerificationCodeSpec()
    {
        Query
            .Where(vc => vc.DateUsed == null)
            .Where(vc => vc.DateCreated.AddMinutes(30) > DateTime.Now);
    }
}
```
The verification code object has a method IsValid(). This shares the same logic used in the ValidVerificationCodeSpec.
Right now I can reuse the code but it reads difficult.

``` c#
public bool IsValid()
{
    return new ValidVerificationCodeSpec().Evaluate(new[] { this }).Any();
}
```
It would read much easier if I just can say `.Evaluate(this)`.
``` c#
public bool IsValid()
{
    return new ValidVerificationCodeSpec().Evaluate(this);
}
```
If the business requirements ever change I don't have to worry about other places where I need to implement the change.

I'd love to get some feedback on this.

Some things still to consider.

- It only makes sense for where clauses to evaluate against one entity. The developer can still make a mistake by using the wrong specification.

### DIFF
--- a/ArdalisSpecification/src/Ardalis.Specification/Evaluators/IInMemorySpecificationEvaluator.cs
+++ b/ArdalisSpecification/src/Ardalis.Specification/Evaluators/IInMemorySpecificationEvaluator.cs
@@ -9,5 +9,6 @@ namespace Ardalis.Specification
     {
         IEnumerable<TResult> Evaluate<T, TResult>(IEnumerable<T> source, ISpecification<T, TResult> specification);
         IEnumerable<T> Evaluate<T>(IEnumerable<T> source, ISpecification<T> specification);
+        bool Evaluate<T>(T source, ISpecification<T> specification);
     }
 }
--- a/ArdalisSpecification/src/Ardalis.Specification/Evaluators/InMemorySpecificationEvaluator.cs
+++ b/ArdalisSpecification/src/Ardalis.Specification/Evaluators/InMemorySpecificationEvaluator.cs
@@ -56,5 +56,10 @@ namespace Ardalis.Specification
                 ? source
                 : specification.PostProcessingAction(source);
         }
+
+        public bool Evaluate<T>(T source, ISpecification<T> specification)
+        {
+            return Evaluate(new[] { source }, specification).Any();
+        }
     }
 }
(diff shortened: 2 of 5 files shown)


fiseni:
Hey @JelmarVA,

Thank you for the feedback. 
The specification pattern originally was created exactly for this purpose, as a mechanism of combining and encapsulating set of business rules. So, your idea is great and you're on the right track.

Indeed, our plan is to cover that "validation" part as well, not just evaluating the specs for querying the persistence. We gave a thought to this, but the reason that we didn't include the possibility of evaluating a single Entity is that the API is not ready for it yet. The design of the specification and the builder is focused more on "filtering" rather than "validating". Imagine you have applied ordering, pagination; it creates ambiguity, right? It shouldn't contain collection-focused functionalities (not even Where).

I was thinking of this, and we might introduce a new type of specification, or perhaps just a specific builder for this purpose. Conceptually, all we need is the entity as a parameter, and bool as a result, right? So, we might hold `Func<T, boo>` delegates.
We can have something as follows:

```c#
public class ValidVerificationCodeSpec : Specification<VerificationCode>
{
    public ValidVerificationCodeSpec()
    {
        Validators.Add(entity =>
        {
            if (entity.DateUsed == null && entity.DateCreated.AddMinutes(30) > DateTime.Now) return true;

            return false;
        }
    }
}
```

And the specification might expose `IsValid` method as follows

```c#
	
public bool IsValid(T entity)
{
    bool result = false;
    foreach (var validator in specification.validators)
    {
        result = result && validator(entity);
    }

    return result;
}
```

This is still not refined, but it might go along with this idea. If you're interested, you can work on this idea.
You can open an issue, and we can track this feature. Once we all agree on the API, we'll implement it. Sounds OK?

JelmarVA:
@fiseni I created an Issue #111 
I'd love to give it a shot to work on this idea.

fiseni:
Thanks, we'll continue the discussion there.
Sure, we're open to any feedback/idea, and any contribution is welcomed. Once we reach a consensus on various features or the API, you're welcome to work on it.

PR #115, opened 2021-04-06
Select doc page

ilyanaDev:
this PR adds a doc for Select and also adds its usage to the FAQ

### DIFF
--- a/docs/features/select.md
+++ b/docs/features/select.md
@@ -16,3 +16,15 @@ The `Select` feature defined in Specification behaves the same as `Select` in Li
 ```csharp
 Query.Select(x => x.Name);
 ```
+
+Since this query is now returning a different type, the type of `Name`, rather than of `x`, the base class of the Specification will need to reflect this. Instead of being a `Specification<T>`, the Specification will need to be a `Specification<T, TReturn>`:
+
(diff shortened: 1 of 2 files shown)


docs/features/select.md
```diff
@@ -0,0 +1,18 @@
...
+
+# Select
+
+The `Select` feature defined in Specification behaves the same as `Select` in Linq, and it takes in `IEnumerable<TSource>` and `Func<TSource, TResult>` as its parameters.
+
+`Select` is used to transform elements in a sequence into a new form. In Specification, `Select` is most commonly used to select a single property of each object in a list being queried. For example, the below expression could be used to retrieve only the name of each object:
+
+```csharp
+Query.Select(x => x.Name);
+```
```
ardalis, line 18: Add that when using `Select` you need to change the base class of specification you use, since it needs to include the returned type (I believe). So instead of `Specification<T>` it becomes `Specification<T,TReturn>` or something like that.

ardalis (changes requested):
Looks good - minor addition requested.

ardalis (approved):

ardalis:
Thanks!

PR #123, opened 2021-05-14
Multi-target with .NET Standard 2.0

MisinformedDNA:


fiseni:
Hey @MisinformedDNA,

Great! Multi-targeting actually it's a better idea than publishing a separate package for EF Core 3.

Can you think of any scenario that we can mess it up for someone? 🤔 

- Projects with `net5.0` TFM, will pick up the `netstandard2.1` version. So they will be on EF Core 5. It's OK.
- Projects on NETFX, will pick up the `netstandard2.0` version. So they will be on EF Core 3. That will be OK too.
- Projects with `net3.1` TFM and older? I assume they will pick up the `netstandard2.0`. That's OK perhaps? 🤔 Anyhow, right now they can't consume the `Ardalis.Specification.EntityFrameworkCore v5` since it's on 2.1, right?

MisinformedDNA:
@fiseni I don't see any downsides.

fiseni (approved):

ardalis (approved):

Huacnin:
I have just started using Ardalis and I find its architecture interesting for developing applications with clean code.
I am developing an app that has three entities: Category, Document, Product and everything works fine except when I want to list the products where the name of the category and the document appear.
This is my Product entity
`namespace Domain.Entitys
{
    public class Product: AuditableBaseEntity
    {
        public int IdCategory {get; set; }
        public int IdDocument {get; set; }
        public string Code {get; set; }
        public string Name {get; set; }
        public string UM {get; set; }
        public decimal PurchasePrice {get; set; }
        public decimal SalesPrice {get; set; }
        public int Stock {get; set; }

        public Category Categories {get; set; }
        public Document Documents {get; set; }
    }
}`
and my ProductDto:
`public class ProductDto
    {
        public int Id {get; set; }
        
        public int IdCategory {get; set; }

        public string Category {get; set; }

        public int IdDocument {get; set; }

        public string Document {get; set; }

        public string Code {get; set; }

        public string Name {get; set; }

        public string UM {get; set; }

        public decimal PurchasePrice {get; set; }

       public decimal SalesPrice {get; set; } 
        
        public int Stock {get; set; }

        public bool Status {get; set; }`

When making the list I get this:

    "pageNumber": 1,
    "pageSize": 10,
    "succeeded": true,
    "message": null,
    "errors": null,
    "data": [
        {
            "id": 4,
            "idCategory": 1,
            "category": null,
            "idDocument": 1,
            "document": null,
            "codigo": "CPMK0004",
            "nombre": "Desayuno Cumpleaños",
            "um": "Pack",
            "precioCompra": 15.20,
            "precioVenta": 21.50,
            "stock": 10,
            "estado": true
        },

My PagesProduct is as follows:
`public PageProducts (int pageSize, int pageNumber, string name, string code)
        {
            Query.Skip ((pageNumber -1) * pageSize)
            .Take (pageSize);

            Query.Include (x => x.Categorys)
                    .Where (x => x. Status == true)
                    .OrderByDescending (x => x.Id);

            if (! string.IsNullOrEmpty (name))
            {
                Query.Search (x => x.Name, "%" + name + "%");
            }

            if (! string.IsNullOrEmpty (code))
            {
                Query.Search (x => x.Code, "%" + code + "%");
            }
        }`

They could tell me where I may be failing so that the list does not match the names of those properties.
Thanks for your support.
Greetings from Peru.

MisinformedDNA:
@Huacnin Please file a new issue.

Issue #112, opened 2021-03-23
Add build and publish tasks for EF6 package.


fiseni:
Actions to be taken:

- Separate build task for EF6 projects (package and test). Since the projects target `.NET Framework v4.7.2`, probably we won't be able to use Ubuntu host and we'll need MSBuild here.
- Task for publishing to Nuget. Right now, no package is generated on build. I can't remember what was the process in `.NET Framework` and in the old .csproj for this (I just know it was not easy :grinning:). Also have to add the package information (packageid, name, author, etc.)

fiseni:
It's fixed. Also, added build and test task for EF6 package. Pushed to Nuget.

PR #142, opened 2021-08-26
Create EF6 package

### LINKED ISSUE #124: Publish EF6 Package to NuGet
Source is here:
https://github.com/ardalis/Specification/tree/main/ArdalisSpecificationEF/src/Ardalis.Specification.EntityFramework6

MisinformedDNA:
Fixes #124

### DIFF
--- a/ArdalisSpecificationEF/src/Ardalis.Specification.EntityFramework6/Ardalis.Specification.EntityFramework6.csproj
+++ b/ArdalisSpecificationEF/src/Ardalis.Specification.EntityFramework6/Ardalis.Specification.EntityFramework6.csproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <TargetFramework>net472</TargetFramework>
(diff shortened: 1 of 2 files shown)


ngruson:
> That's great @MisinformedDNA. Thank you!
> 
> So, actually, we can use SDK-style projects? Why we've been bothering all this time then? 😅 Now that I'm thinking, most of my libraries that support 4.x are SDK style too. Brain freeze.
> 
> @ngruson You were helping us with EF6 and you created the project. Do you have any reason why we shouldn't use SDK style projects? Do you have any specific use-case?
> Btw, sorry that you have been waiting for so long. We simply got stuck with the legacy structure and never published it.

No specific reason not to use SDK style projects. At the time, I just created a new .NET Framework 4.8 class library in VS2019 which creates a non-SDK style project by default.

fiseni:
Thanks Nils. The package is generated and I see no issues. 
But, the publish task is failing since we need .NET Framework runtime. I see I have an issue on that one #112, and probably we'll need a Windows environment. We'll check how to do that with Github Actions.

PR #146, opened 2021-09-02
Refactor the solution and folder structure.

fiseni:
This PR includes:
- Refactoring of solution and folder structure.
- Updated the paths in build and publish tasks.
- Updated the paths in docker files.

<strong>Solution structure:</strong>

SpecificationSolutionStructure

<strong>Folder structure:</strong>

SpecificationFolderStructure


MisinformedDNA:
I would rename the file folder names `SpecificationEF6` and `SpecificationEFCore` to match the solution folder names `Specification.EntityFramework6` and `Specification.EntityFrameworkCore`. This would also be more consistent with the package names.

fiseni:
I started that way, but the path is getting too long. One of the projects hit 256 chars I think. So, it's a little compromise here :)

fiseni:
Ah, what the heck, let's do it that way. I was trying to be considerate for users, they might place the project folder in a nested location in their environment. But, it's better to be consistent.

ardalis (approved):

PR #153, opened 2021-09-07
Add SQL LIKE implementation for the in-memory evaluator.

### LINKED ISSUE #150: Add SQL LIKE implementation for the in-memory evaluator.
Starting with version 5 of the package, we introduced in-memory evaluators. The usage is as follows:

```c#
List<Customer> customers = new List<Customer>();
var spec = new CustomerByNameSpec("somename");
var result = spec.Evaluate(customers);
```
This greatly simplifies the usage if you need to apply a specification to some collection you already have in hand. Not rarely, in your domain model, you might need to filter a given collection (it might be the navigation of your entity). If that action is required in several places, you may want to encapsulate it in a separate specification, and then just apply it wherever you need it. That's the use-case, generally.
Since the in-memory evaluator operates against in-memory collections, it acts independently from the plugin packages (EFCore, EF6). Therefore, all ORM-specific features that the specification may contain are excluded from this evaluation. The following features won't be evaluated:
- Include chains
- AsNoTracking
- AsSplitQuery
- AsNoTrackingWithIdentityResolution

But, unfortunately, since the `Search` feature is evaluated to SQL `Like` by the plugin packages, it is excluded as well. This is quite inconvenient. If I have applied `Search` in my specification, I'd like to be processed during in-memory evaluation too.

The solution would be to implement the SQL `LIKE` operator in C#. But, `LIKE` is a very rich and powerful operator with plenty of options, and trying to match and mimic the behavior is not a simple task.

I'm open to ideas here. Does anyone know of such an implementation?
My first thought was to check how that is implemented in `EntityFrameworkCore.SQLite`. This [SO post](https://stackoverflow.com/a/22377898/10577116) might also be helpful

fiseni:
I think this will do the trick. Check the related issue for more details.
Fixes #150


fiseni:
@ardalis FYI.
If you're ok here, let's merge this one (as long there are no merge conflicts 😄 )

ardalis (approved):

PR #157, opened 2021-09-15
Add dependency injection for Repository to Getting Started guide

davidhenley:


fiseni:
Thanks, @davidhenley,
This page was a contribution by a community member. I see the entity is named "Herstellerkontakt" (perhaps it was some copy-paste from his own code). Would you mind renaming that to "Customer"?

Also, usually, most of the users would use interfaces, and not inject directly the repository implementation. But, we can fix that in a separate PR.

davidhenley:
Sure thing. Done!

Issue #162, opened 2021-09-25
Add example usage on how to extend the specification builder with Caching


vittorelli:
As suggested in #161. I'll create a PR for updating the docs

PR #163, opened 2021-09-25
Docs builder extensions #162

vittorelli:
Work in progress. Does this look okay?  
Or should the docs contain less code and move that into the sample projects in stead?

### DIFF
--- a/docs/extensions/create-specification-builder.md
+++ b/docs/extensions/create-specification-builder.md
@@ -8,7 +8,17 @@ nav_order: 2
 
 # How to create your own specification builder
 
-How to create your own specification builder 
+The specification builder from `Ardalis.Specification` is extensible by design. In fact, the methods you can use out of the box are implemented as extension methods themselves (check out the [source code](https://github.com/ardalis/Specification/blob/main/Specification/src/Ardalis.Specification/Builder/SpecificationBuilderExtensions.cs)). Your project might have requirements that cannot be satisfied by the existing toolset of course, or you might want to simplify repetitive code in several specification constructors. Whatever your case, enhancing the default builder is easy by creating your own extension methods.
+
+So where do you start? A good practice is to write the thing you think you need. Say you'd like to use a builder method `WithCustomerIdAndName` that takes the `Id` and `Name` of a customer as parameters. Then just write it like so:
+
+````csharp
+Query.AsNoTracking()
+    .WithCustomerIdAndName(1337, "John Doe");
+````
+
+From here you can inspect the return type of the builder method you chained it to (`AsNoTracking`), and create an extension method on that interface (it doesn't need to be chained of course -- working on `Query` itself is also valid). This will most likely be `ISpecificationBuilder<T>`, but in some cases it's an inherited inteface. The example below illustrates how extension methods on inherited interfaces allow the builder to offer specific methods in specific contexts.
+
 
 ## Example: Configure caching behaviour through specification builder extension method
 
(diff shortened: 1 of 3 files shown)


docs/extensions/create-specification-builder.md
```diff
@@ -5,4 +5,70 @@ parent: Extensions
 nav_order: 2
 ---
 
-How to create your own specification builder
+
+# How to create your own specification builder
+How to create your own specification builder
```
ardalis, line 10:
```suggestion

How to create your own specification builder 
```

ardalis: Also, although the title is all about "How" it would be great to say "Why" one would want/need to do this somewhere at the top, too.

docs/extensions/create-specification-builder.md
```diff
@@ -5,4 +5,70 @@ parent: Extensions
...
 ---
 
-How to create your own specification builder
+
+# How to create your own specification builder
+How to create your own specification builder 
+
+## Example: Configure caching behaviour through specification builder extension method
+
+In order to achieve this:
```
ardalis, line 14:
```suggestion
In order to achieve this (note the `.WithTimeToLive` method):
```

docs/extensions/create-specification-builder.md
```diff
@@ -5,4 +5,70 @@ parent: Extensions
...
+}
+````
+
+This extension method can only be called when chained after `SpecificationBuilderExtensions.EnableCache`. This is because `EnableCache` returns `ICacheSpecificationBuilder<T>` which inherits from `ISpecificationBuilder<T>`.
+
+```csharp
+// TODO: Repository example
+```
+
+Finally, we need to take of some plumbing to implement both `` and ``. The class below uses `ConditionalWeakTable` to do the trick. An other solution is to create a base class that inherits from `Specification<T>`.
```
ardalis, line 50:
```suggestion
Finally, we need to take of some plumbing to implement both `` and ``. The class below uses `ConditionalWeakTable` to do the trick. Another solution is to create a base class that inherits from `Specification<T>`.
```

docs/extensions/create-specification-builder.md
```diff
@@ -5,4 +5,70 @@ parent: Extensions
...
+
+````csharp
+public static class SpecificationExtentions
```
ardalis, line 53:
```suggestion
public static class SpecificationExtensions
```

ardalis (changes requested):
Looks great - having code in the docs is good so folks don't need to hunt down the sample separately. You can link to the sample as needed if it's something where seeing it in context of a whole app will add clarity.

fiseni:
You're doing great work here @vittorelli.
While designing the infrastructure and the multi-level chain structure, never thought anyone will analyze it thoroughly or make use of it 😄 I'm glad that users are utilizing this extensibility.

vittorelli:
> Looks great - having code in the docs is good so folks don't need to hunt down the sample separately. You can link to the sample as needed if it's something where seeing it in context of a whole app will add clarity.

> Also, although the title is all about "How" it would be great to say "Why" one would want/need to do this somewhere at the top, too.

I've updated the docs so it answers to the "Why". I didn't add it as a sample in the source because I felt it would somewhat conflict with the existing `CachedRepository` sample.

vittorelli:
> You're doing great work here @vittorelli.
> While designing the infrastructure and the multi-level chain structure, never thought anyone will analyze it thoroughly or make use of it 😄 I'm glad that users are utilizing this extensibility.

@fiseni Thanks :-)  
I'm happy to contribute. It helped me a lot on several projects!

ardalis (approved):

PR #191, opened 2021-12-21
Implement IgnoreQueryFilters feature.

fiseni:
Closes #159

PR #192, opened 2021-12-21
Add support for condition in the specification builder methods.

### LINKED ISSUE #143: Can multi-condition query add WhereIf extension similar to abp
When there are multiple query conditions at the same time and do not know whether each query condition has a value, we must first determine whether the condition is empty, and then determine whether to add it to the filter condition, using Where is more troublesome, whether Provide a WhereIf extension similar to Abp, that is very convenient, this is just a request for optimization, thank you.

```
db.Employee
.WhereIf<Employee>(!string.IsNullOrEmpty(query.key), n => n.Name == query.key || n.Description.Contains(query.key)).ToList();
```

fiseni:
Closes #143. This should be merged after PR #191.

All specification builder methods got an overload that accepts `bool condition` parameter. If this condition is false, the given expression/value is discarded and is not added to the specification. I decided not to implement it as an optional parameter, since that might be a breaking change for someone. The `Select` and `PostProcessingAction` methods are an exception, and those methods won't have an overload with this parameter. Once the specification is defined with generic parameters`<T, TResult>`, then it must contain a selector.

Sample Usage:
```c#
public class CustomerSpec : Specification<Customer>
{
    // Instead of having this
    public CustomerSpec(CustomerFilter filter)
    {
        if (!string.IsNullOrEmpty(filter.Name))
            Query.Where(x => x.Name == filter.Name);

        if (!string.IsNullOrEmpty(filter.Email))
            Query.Where(x => x.Email == filter.Email);

        if (!string.IsNullOrEmpty(filter.Address))
            Query.Search(x => x.Address, "%" + filter.Address + "%");
    }

    // Users can do this
    public CustomerSpec(CustomerFilter filter)
    {
        Query
            .Where(x => x.Name == filter.Name, !string.IsNullOrEmpty(filter.Name))
            .Where(x => x.Email == filter.Email, !string.IsNullOrEmpty(filter.Email))
            .Search(x => x.Address, "%" + filter.Address + "%", !string.IsNullOrEmpty(filter.Address));
    }
}
```

Since the `Include` and `Order` methods offer multi-level chaining, if the parent method has false condition, the whole chain afterward is discarded.

```c#
public class CompanyByIdWithFalseConditions : Specification<Company>, ISingleResultSpecification
{
    public CompanyByIdWithFalseConditions(int id)
    {
        Query.Where(x => x.Id == id, false)
            .OrderBy(x => x.Id, false)
                .ThenBy(x => x.Name) // since the parent is not added, this also will be discarded
            .Include(x => x.Stores, false)
                .ThenInclude(x => x.Products); // since the parent is not added, this also will be discarded
    }
}

public class CompanyByIdWithFalseConditionsForInnerChains : Specification<Company>, ISingleResultSpecification
{
    public CompanyByIdWithFalseConditionsForInnerChains(int id)
    {
        Query.Where(x => x.Id == id, false)
            .OrderBy(x => x.Id) // it will be added
                .ThenBy(x => x.Name, false) // it won't be added
                .ThenByDescending(x => x.Name) // since the parent is not added, this also will be discarded
            .Include(x => x.Stores) // it will be added
                .ThenInclude(x => x.Products, false) // it won't be added
                .ThenInclude(x => x.Origin) // since the parent is not added, this also will be discarded
    }
}
```

### DIFF
--- a/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
+++ b/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
@@ -56,9 +56,9 @@ namespace Ardalis.Specification.EntityFramework6
             await SaveChangesAsync(cancellationToken);
         }
         /// <inheritdoc/>
-        public virtual async Task SaveChangesAsync(CancellationToken cancellationToken = default)
+        public virtual async Task<int> SaveChangesAsync(CancellationToken cancellationToken = default)
         {
-            await dbContext.SaveChangesAsync(cancellationToken);
+            return await dbContext.SaveChangesAsync(cancellationToken);
         }
 
         /// <inheritdoc/>
--- a/Specification.EntityFrameworkCore/src/Ardalis.Specification.EntityFrameworkCore/Evaluators/IgnoreQueryFiltersEvaluator.cs
+++ b/Specification.EntityFrameworkCore/src/Ardalis.Specification.EntityFrameworkCore/Evaluators/IgnoreQueryFiltersEvaluator.cs
@@ -1,0 +1,26 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Ardalis.Specification.EntityFrameworkCore
+{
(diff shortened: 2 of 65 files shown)


ardalis:
Add to docs. 2nd parameter, if false, means the expression will not be added.

PR #193, opened 2021-12-22
Implement specification validators.

### LINKED ISSUE #111: Evaluate Specification for one Entity
It would be nice to use a specification to make sure an object is still valid.
I've given[ an example](https://github.com/ardalis/Specification/pull/110#issue-595615036) of why you'd want to share this logic in one place.

@fiseni proposed the [following](https://github.com/ardalis/Specification/pull/110#issuecomment-802019868):
> I was thinking of this, and we might introduce a new type of specification, or perhaps just a specific builder for this purpose. Conceptually, all we need is the entity as a parameter, and bool as a result, right? So, we might hold `Func<T, bool>` delegates.
> We can have something as follows:
> 
> ```cs
> public class ValidVerificationCodeSpec : Specification<VerificationCode>
> {
>     public ValidVerificationCodeSpec()
>     {
>         Validators.Add(entity =>
>         {
>             if (entity.DateUsed == null && entity.DateCreated.AddMinutes(30) > DateTime.Now) return true;
> 
>             return false;
>         }
>     }
> }
> ```
> 
> And the specification might expose `IsValid` method as follows
> 
> ```cs
> 	
> public bool IsValid(T entity)
> {
>     bool result = true;
>     foreach (var validator in specification.validators)
>     {
>         result = result && validator(entity);
>     }
> 
>     return result;
> }
> ```
> 
> This is still not refined, but it might go along with this idea.

Let's say we add a Validator builder. This should be translatable to a query on the database right?
Or do you guys see it so they live together next to each other as following?  
```cs
public class ValidVerificationCodeSpec : Specification<VerificationCode>
{
    public ValidVerificationCodeSpec()
    {
        Query
            .Where(vc => vc.DateUsed == null)
            .Where(vc => vc.DateCreated.AddMinutes(30) > DateTime.Now);

        Validators.Add(entity =>
        {
            if (entity.DateUsed == null && entity.DateCreated.AddMinutes(30) > DateTime.Now) return true;

            return false;
        }
    }
}
```
That will result in some duplication of the logic.

fiseni:
Closes #111. This should be merged after PR #192.

This feature offers the possibility to check if an entity satisfies the conditions in a given specification. Since it's a single entity, it's being validated only against expressions/conditions specified in `Where` and `Search`, and everything else is omitted during this process. The usage is quite simple

```c#
var customer = new Customer { Name = "something" };
var spec = new CustomerByNameSpec("something");
var result = spec.IsSatisfiedBy(customer)
```

The implementation uses a separate infrastructure for this feature, by introducing new constructs `ISpecificationValidator` and `IValidator` (partial validator). I decided not to use the in-memory evaluators, since we'll force omitted evaluators to implement `IsValid()`, which is not nice.

### DIFF
--- a/Specification.EntityFramework6/tests/Ardalis.Specification.EntityFramework6.IntegrationTests/RepositoryOfT_ListAsync.cs
+++ b/Specification.EntityFramework6/tests/Ardalis.Specification.EntityFramework6.IntegrationTests/RepositoryOfT_ListAsync.cs
@@ -167,7 +167,7 @@ namespace Ardalis.Specification.EntityFramework6.IntegrationTests
 
             result.Should().NotBeNull();
             result.Should().ContainSingle();
-            result[0].Id.Should().Be(StoreSeed.VALID_Search_City_ID);
+            result[0].Id.Should().Be(StoreSeed.VALID_Search_ID);
             result[0].City.Should().Contain(StoreSeed.VALID_Search_City_Key);
         }
     }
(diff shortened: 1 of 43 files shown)


ardalis:
Needs docs.

ardalis (approved):

PR #196, opened 2021-12-28
Restore ISingleResultSpecificationOfT

vittorelli:
The ISingleResultSpecificationOfT seems to be have been removed in 73693ac1b7ada2ac889c89c1e1b7882f19f566cb while being included in the current Nuget package. I assume this was accidental?

~~Furthermore, I've added two base classes SingleResultSpecification based on the existing Specification classes.~~


vittorelli:
I'm having second thoughts on the SingleResultSpecification classes... This does create the situation where users can define single result specifications multiple ways and may lead to code inconsistencies. On the other hand, it does conform to the existing Specification classes.

What do you guys think?

devbased:
I don't understand what is the point of `ISingleResultSpecification`? What does it provide?
Here i've removed it
```csharp
public interface IReadRepositoryBase<T> where T : class
{
    // other methods
    Task<T?> GetBySpecAsync(ISpecification<T> specification, CancellationToken cancellationToken = default);
}
public abstract class RepositoryBase<T> : IRepositoryBase<T> where T : class
{
    // other methods
    public virtual async Task<T?> GetBySpecAsync(ISpecification<T> specification, CancellationToken cancellationToken = default)
    {
        return await ApplySpecification(specification).FirstOrDefaultAsync(cancellationToken);
    }
}
```

vittorelli:
It's a marker interface to indicate that a specification returns a single result (in LINQ: `SingleOrDefault`, `FirstOrDefault`, `Single` and `First`). For me, it has two advantages:

 * Help the developer understand the intent of a specification at a glimpse
 * Allow the IDE to identify incorrect usage of specifications.

For the first point, consider the example below. Using the `ISingleResultSpecification` a developers easily understands that the first specification returns multiple customers while the latter returns a single one -- even without having to read the details in the constructor (parameters, body) or fully analysing the class name.

````csharp
// The first word in this specification is plural, to indicate multiple customers are returned
public class CustomersWithAddressesAndPurchaseHistory : Specification<Customer>
{
    public CustomersWithAddressesAndPurchaseHistoryIncluded(bool includeAddress, bool includePurchaseHistory)
    { 
        // Query logic
    }
}

// The first word in the specification name is singular, but this is hard to detect.
// The ISingleResultSpecification marker interface is more obvious, which helpt to understand the intent.
public class CustomerWithAddressesAndPurchaseHistory : Specification<Customer>, ISingleResultSpecification
{
    public CustomerWithAddressesAndPurchaseHistoryIncluded(int id, bool includeAddress, bool includePurchaseHistory)
    { 
        // Query logic
    }
}
````

For the second point, consider the example below. When used in combination with carefully designed repository interfaces, one can generate compile time errors for when a developer accidentally uses the wrong specification. Imagine you have two specification with similar names and untrivial constructor parameters, where one returns a list of entities and the other a single entity. We can then use the compiler to generate errors if one accidentally picked the wrong specification.

````csharp
    interface IRepository<T>
    {
        T ListAsync(ISpecification<T> spec);

        T SingleAsync<Spec>(Spec spec)
            where Spec : ISingleResultSpecification, ISpecification<T>;
    }
````

The generic `ISingleResultSpecification<T>` interface is included in version 5.2.0 NuGet package; removing it in future package creates an unannounced breaking change. I haven't modified my pull request, but my suggestion is to mark the generic interface as obsolete, leaving only `ISingleResultSpecification`.

vittorelli:
The generic `ISingleResultSpecification<T>` interface is included in version 5.2.0 NuGet package; removing it in future package creates an unannounced breaking change. I haven't modified my pull request, but my suggestion is to mark the generic interface as obsolete, leaving only `ISingleResultSpecification`.

devbased:
@vittorelli Thank you for explanation. But is it really a specification responsibility to define resulting object? Would you create marker interfaces for TwoResults, ThreeResults, NResults, IAsyncEnumerableResult or ListResult, or other complex type specifications? If so, should ISomeResultSpecification create that result and not just mark that it 'wants' it? IMHO Specification builds query and not materializes it (or tells you how to materialize).

Sorry if this is not a right place for such discussion 😅

fiseni:
Hi,

The `ISingleResultSpecification` interface has no importance in the specification infrastructure, nor it affects how they're built. Ultimately, it's utilized by the repository implementation and offers help if you need to restrict how the way a given specification can be materialized. I personally am not using it. Not rarely my specs may contain several constructors (accepting Id, collection of Ids, and possibly a filter object), so this interface does not make sense for my use-cases. But, the community asked for it, so it's a thing now. I do admit, including the interface out of the box and restricting the `GetBySpec` repository method was an opinionated action. As I said, it's too late now, and taking it away will be worse. Anyhow, the built-in repository implementation is just for reference, users can have their own implementation or use the specs without repositories (directly with dbContext). Also, some users have created interesting usages around this interface, so we'll keep it. @vittorelli  thank you for spotting this, it seems is being removed by mistake.

As for the additional base class `SingleResultSpecification`, I'm not for it honestly. Not sure what @ardalis thinks about this, but I think libraries shouldn't expose too many "notions". In our case, all that users have to remember is that they need to extend the base `Specification` class and use `Query` property to start building the specification. Once we include new "starting points", we introduce dilemmas somehow, should include more detailed clarifications, etc. Users can build their own infrastructure on top of the base one provided in the library. It's very common to have project-specific base specifications, which may include some generic actions like pagination. It's up to the users.

vittorelli:
Ok, I agree. Better to keep the package clean and leave it up to end users if they want to use `ISingleResultSpecification` and/or create SingleResultSpecification base classes :-)

I've updated the PR so it only restores the generic interface so as to not have breaking changes in the next NuGet release.

fiseni:
Thank you @vittorelli. We appreciate your help.

Issue #202, opened 2022-01-10
Docs: How to use Specification without a repository


ardalis:
Refer to this issue for syntax:
https://github.com/ardalis/Specification/issues/86

ardalis:
Fixed by #204

Issue #201, opened 2022-01-10
Docs: How to use Specifications with the Repository Pattern


ardalis:
Needs completed.
https://ardalis.github.io/Specification/usage/use-specification-repository-pattern.html

Various implementations shown in closed issues, as well as in the eShopOnWeb and CleanArchitecture solutions.

ardalis:
@kylemcmaster

ardalis:
Fixed by #204

PR #204, opened 2022-01-11
Update docs for use with DbContext and Repository Pattern

KyleMcMaster:
- adding documentation to satisfy #201
- adding documentation to patially satisfy #202 for using Specifications with DbContexts
- cleanup using unnecessary statements found while navigating code

### DIFF
--- a/docs/usage/use-specification-dbcontext.md
+++ b/docs/usage/use-specification-dbcontext.md
@@ -8,5 +8,46 @@ nav_order: 3
 # How to use Specifications with a DbContext
 
 You can use Specifications to define queries that are executed directly using an EF6 or EF Core `DbContext`.
+The following snippet defines a `Customer` entity and a sample `DbContext` which defines a `DbSet` of Customers.
 
-(todo)
+```csharp
+public class Customer
+{
+    public int Id { get; set; }
+    public string Name { get; set; }
+}
+
+public class SampleDbContext : DbContext
+{
+    public DbSet<Customer> Customers { get; set; }
+
+    public SampleDbContext(DbContextOptions<SampleDbContext> options)
+        : base(options)
+    {
+    }
+
+    // additional overrides intentionally left out
+}
+```
+
+A specification can be applied directly to this `DbSet` using the `WithSpecification` extension method defined in `Ardalis.Specification.EntityFrameworkCore` package. Assuming a Specification is defined similar to the `ItemByIdSpec` described in [How to Create Specifications](./creating-specifications.md), the following code demonstrates putting these pieces together.
+
+```csharp
+int id = 1;
+
+var specification = new CustomerByIdSpec(id);
+
+var customer = dbContext.Customers
+    .WithSpecification(specification)
+    .FirstOrDefault();
+```
+
+Note that the `WithSpecification` extension method exposes an `IQueryable<T>` so additional extension methods maybe be applied after the specification. Some examples can be found below.
+
+```csharp
+bool isFound = dbContext.Customers.WithSpecification(specification).Any();
+
+int customerCount = dbContext.Customers.WithSpecification(specification).Count();
+
+var customers = dbContext.Customers.WithSpecification(specification).ToList();
+```
--- a/docs/usage/use-specification-dbcontext.md
+++ b/docs/usage/use-specification-dbcontext.md
@@ -33,6 +33,8 @@ public class SampleDbContext : DbContext
 A specification can be applied directly to this `DbSet` using the `WithSpecification` extension method defined in `Ardalis.Specification.EntityFrameworkCore` package. Assuming a Specification is defined similar to the `ItemByIdSpec` described in [How to Create Specifications](./creating-specifications.md), the following code demonstrates putting these pieces together.
 
 ```csharp
+// handling of IDisposable DbContext intentionally left out
+
 int id = 1;
 
 var specification = new CustomerByIdSpec(id);
--- a/docs/usage/use-specification-repository-pattern.md
+++ b/docs/usage/use-specification-repository-pattern.md
@@ -7,7 +7,7 @@ nav_order: 2
 
 # How to use Specifications with the Repository Pattern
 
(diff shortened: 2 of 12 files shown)


docs/usage/use-specification-dbcontext.md
```diff
@@ -8,5 +8,46 @@ nav_order: 3
...
+
+A specification can be applied directly to this `DbSet` using the `WithSpecification` extension method defined in `Ardalis.Specification.EntityFrameworkCore` package. Assuming a Specification is defined similar to the `ItemByIdSpec` described in [How to Create Specifications](./creating-specifications.md), the following code demonstrates putting these pieces together.
+
+```csharp
+int id = 1;
+
+var specification = new CustomerByIdSpec(id);
+
+var customer = dbContext.Customers
+    .WithSpecification(specification)
```
ardalis, line 41: Since this will only work if the required `using` statement is available can we make note of that or include it in the listing somehow? A comment at the top is probably sufficient.

docs/usage/use-specification-repository-pattern.md
```diff
@@ -7,6 +7,61 @@ nav_order: 2
 
 # How to use Specifications with the Repository Pattern
 
-Specifications shine when combined with the [Repository Pattern](https://deviq.com/design-patterns/repository-pattern), a sample generic implementation of which is included in this NuGet package.
+Specifications shine when combined with the [Repository Pattern](https://deviq.com/design-patterns/repository-pattern), a sample generic implementation of which is included in this NuGet package. For the purpose of this walkthrough, the repository can be thought of as a simple data access abstraction over a collection of entities. In this example, the entity for a `Hero` and it's repository interface are defined below.
```
ardalis, line 10:
```suggestion
Specifications shine when combined with the [Repository Pattern](https://deviq.com/design-patterns/repository-pattern), a sample generic implementation of which is included in this NuGet package. For the purpose of this walkthrough, the repository can be thought of as a simple data access abstraction over a collection of entities. In this example, the entity for a `Hero` and its repository interface are defined below.
```

docs/usage/use-specification-repository-pattern.md
```diff
@@ -7,6 +7,61 @@ nav_order: 2
...
+}
+
+public interface IHeroRepository
+{
+    List<Hero> GetAllHeroes();
+}
+
+```
+
+A Specification can be defined for the `Hero` entity when filtering is required for the collection of all heroes. Note any fields that are needed to filter the heroes are passed to the Specification's constructor where the query logic should be implemented. In this case, a Specification is defined that filters heroes by whether the hero is alive and is an Avenger.
```
ardalis, line 28:
```suggestion
A Specification can be defined for the `Hero` entity to allow filtering the collection of heroes. Note any fields that are needed to filter the heroes are passed to the Specification's constructor where the query logic should be implemented. In this case, a Specification is defined that filters heroes by whether the hero is alive and is an Avenger.
```

docs/usage/use-specification-repository-pattern.md
```diff
@@ -7,6 +7,61 @@ nav_order: 2
...
+```
+
+With the Specification and Repository defined, it is now possible to define a `GetHeroes` method that can take a `HeroRepository` as a parameter along with the filtering conditions and produce a filtered collection of heroes. Applying the Repository to the Specification is done using the `Evaluate` method on the Specification class which takes a `IEnumerable<T>` as a parameter. This should mirror the kind of methods typically found on Controllers or [Api Endpoints](https://github.com/ardalis/ApiEndpoints) where the IHeroRepository might be supplied via Dependency Injection to the class's constructor rather than passed as a parameter.
+
+```csharp
+public IEnumerable<Hero> GetHeroes(IHeroRepository repository, bool isAlive, bool isAvenger)
+{
+    var specification = new HeroByIsAliveAndIsAvengerSpec(isAlive, isAvenger);
+
+    return specification.Evaluate(repository.GetAllHeroes());
```
ardalis, line 47: We'd only want to use `Evaluate` on existing in-memory collections. I don't think we want to propose that one use it on a repository's results, as that would always be performed in-memory and would be less efficient than filtering the data within the data source itself.

We do need docs and samples of `Evaluate`, but this document should show how one can pass a `Specification` to a repository method and have the repository use that specification to perform querying logic internally. Essentially the code here should be similar to but separate from the built-in repository, for users who want to add support for specifications to their own repository rather than inheriting from the one we provide in this package.

docs/usage/use-specification-repository-pattern.md
```diff
@@ -7,6 +7,61 @@ nav_order: 2
...
+}
+```
+
+Suppose the data store behind the IHeroRepository has the following state and client code calls the `GetHeroes` as below. The value of result should be a collection containing only the Spider Man hero.
```
ardalis, line 51:
```suggestion
Suppose the data store behind the IHeroRepository has the following state and client code calls the `GetHeroes` as below. The result should be a collection containing only the Spider Man hero.
```

ardalis (changes requested):
Great start!

docs/usage/use-specification-dbcontext.md
KyleMcMaster: updated

docs/usage/use-specification-repository-pattern.md
KyleMcMaster: updated

docs/usage/use-specification-repository-pattern.md
KyleMcMaster: updated

docs/usage/use-specification-repository-pattern.md
KyleMcMaster: Updated to show how a basic example of how to modify an existing repository to apply specifications.

docs/usage/use-specification-repository-pattern.md
KyleMcMaster: updated

ardalis (approved):

PR #226, opened 2022-01-18
Adding and Applying EditorConfig

### LINKED ISSUE #224: Enable XML comments.
Now that we have significant coverage of the code with XML comments, we may enable them (without this, the comments won't be packed in the Nuget package, and won't be available to the consumers). We have to add the following line in the project files

```
<DocumentationFile>bin\$(Configuration)\Ardalis.Specification.xml</DocumentationFile>
```

Once we add this, we'll get warnings if we missed adding an XML comment for any publicly visible construct.

ardalis:
Fixes #224

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,126 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+###############################
+# Core EditorConfig Options   #
+###############################
+# All files
+[*]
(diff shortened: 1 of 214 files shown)


ardalis:
Not seeing any XML comments in the produced *.nupkg file so far with this change, @fiseni

fiseni (approved):

PR #253, opened 2022-03-01
Add Read repository base

frankyjquintero:
* Add Read repository base for EF 6 
* Add Read repository base for EF Core

### DIFF
--- a/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/ReadRepositoryBaseOfT.cs
+++ b/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/ReadRepositoryBaseOfT.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data.Entity;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Ardalis.Specification.EntityFramework6
+{
+  /// <inheritdoc/>
+  public abstract class ReadRepositoryBase<T> : IReadRepositoryBase<T> where T : class
+  {
+    private readonly DbContext dbContext;
+    private readonly ISpecificationEvaluator specificationEvaluator;
+
+    public ReadRepositoryBase(DbContext dbContext)
+        : this(dbContext, SpecificationEvaluator.Default)
+    {
+    }
+
+    /// <inheritdoc/>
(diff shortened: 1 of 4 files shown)


frankyjquintero:
@fiseni @ardalis

fiseni:
Hey @frankyjquintero 

Thank you for your contribution. I'm sorry, but I don't think we'll merge this, we don't want to maintain two separate repository implementations. We're providing just the base implementation, and it's up to you to restrict the exposed behavior through your interfaces. In any case, in your application, you will need to create `IRepository` and `Repository` constructs, right? So, you can do the following (this is the standard approach)

```c#
using Ardalis.Specification;
using Ardalis.Specification.EntityFrameworkCore;
using Microsoft.EntityFrameworkCore;
using Microsoft.Extensions.DependencyInjection;

var services = new ServiceCollection();

services.AddDbContext<AppDbContext>(options => options.UseSqlServer("connectionstring"));
services.AddScoped(typeof(IRepository<>), typeof(Repository<>));
services.AddScoped(typeof(IReadRepository<>), typeof(ReadRepository<>));

public interface IReadRepository<T> : IReadRepositoryBase<T> where T : class
{
}

public interface IRepository<T> : IRepositoryBase<T> where T : class, IAggregateRoot
{
}

public class ReadRepository<T> : RepositoryBase<T>, IReadRepository<T> where T : class
{
  public ReadRepository(AppDbContext dbContext) : base(dbContext)
  {
  }
}

public class Repository<T> : RepositoryBase<T>, IRepository<T> where T : class, IAggregateRoot
{
  public Repository(AppDbContext dbContext) : base(dbContext)
  {
  }
}

public class AppDbContext : DbContext
{
}


public interface IAggregateRoot
{
}

```

frankyjquintero:
Totally agree with your suggestions, but the idea is to separate the read-only base class for scenarios where you don't want to create a repository with the savechanges methods included in add, update, delete.

If the starting point is not to keep 2 the current base could inherit from the read-only implementation such as the Ireadrepository.

I send commit with change suggestion, taking into account that code is not repeated

fiseni:
Sorry, we don't feel for merging this. The `RepositoryBase` is just an implementation. Your read/write implementations can both inherit this one, it doesn't matter if it includes all methods. You will control what is exposed through your interfaces. You will consume the repositories through the interfaces.

Anyhow, the repositories in the library are provided just as a sample and for convenience. You can have your own implementations and design them as you wish (that's what I do, not using the built-in ones).

PR #273, opened 2022-06-20
Add WithSpecification overload for specifications with Select.

### LINKED ISSUE #268: Unable to use WithSpecification with Specification<S,R> that is the result of Select
If I use `Select` as in the following specification I am unable to use `WithSpecification` because the types cannot be inferred. 

```C#
// Note: this is a very basic example
public sealed class AToBSpec : Specification<A, B> {     
	public AToBSpec()     {  
		Query.Select(a => new B()  {}).AsNoTracking();     
  	}
 }
```

And then attempt to use it as follows
```C#
List<B> b = context.A.WithSpecification(new AToBSpec()).ToList();
```

I receive the following error:
```
Cannot convert source type 'System.Collections.Generic.List<A>' to target type 'System.Collections.Generic.List<B>
```

If I include the following additional extension for `WithSpecification` I can make this work (well it compiles, I have not fully tested this yet). Does it make sense to include an override of `WithSpecification` that builds on `Specification<S,R>` so that you can use `Select`?
```C#
public static class DbSetExtensions {     
	public static IQueryable<TRSource> WithSpecification<TSource, TRSource>(this IQueryable<TSource> source, Specification<TSource, TRSource> specification, ISpecificationEvaluator? evaluator = null) where TSource : class     {         
		evaluator = evaluator ?? SpecificationEvaluator.Default;         
		return evaluator.GetQuery(source, specification);    
	} 
}

```

I'm not sure if this is a good solution or if I am just missing something.

fiseni:
Closes #268

### DIFF
--- a/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/Extensions/DbSetExtensions.cs
+++ b/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/Extensions/DbSetExtensions.cs
@@ -8,7 +8,11 @@ namespace Ardalis.Specification.EntityFramework6
 {
   public static class DbSetExtensions
   {
-    public static async Task<List<TSource>> ToListAsync<TSource>(this DbSet<TSource> source, ISpecification<TSource> specification, CancellationToken cancellationToken = default) where TSource : class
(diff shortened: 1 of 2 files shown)


ardalis (approved):
Thanks for getting rid of horizontal scrolling, too :)

PR #312, opened 2023-02-23
Set ef core package version when targeting .net7

Mayis-ST:


fiseni:
Hi @Mayis-ST,

We set only the minimum required EF version. If there are no new features in the EF version that we want to utilize, then there is no benefit in forcing consumers to use the new EF version. For instance, there might be breaking changes in EF, and consumers may want to use `.NET7` with `EF Core 6`.

Anyhow, you're free to use `EF Core 7` with this library. You just need to explicitly install the EF packages, and the dependencies will be updated accordingly.

image

Did you face any issue with it?

Mayis-ST:
I had a very niche issue which only appeared when I was using our company's proprietary make tool.

One of my assemblies had a dependency on the `Ardalis.Specification` package, and also referenced another one of my assemblies that uses `EfCore 7`. Somehow `EfCore` version 6 was being resolved and I was getting some `TypeLoadException`s.

I think the make tool was causing this by selectively building each project, and I fixed the issue by explicitly referencing EfCore 7 from in the first assembly.

PR #316, opened 2023-03-18
Added support for AsAsyncEnumerable

nkz-soft:
Hello @ardalis!
I use AsyncEnumerable in my project, and it would be nice to have it supported in this library.

### DIFF
--- a/Specification.EntityFrameworkCore/src/Ardalis.Specification.EntityFrameworkCore/RepositoryBaseOfT.cs
+++ b/Specification.EntityFrameworkCore/src/Ardalis.Specification.EntityFrameworkCore/RepositoryBaseOfT.cs
@@ -173,6 +173,12 @@ namespace Ardalis.Specification.EntityFrameworkCore
       return await dbContext.Set<T>().AnyAsync(cancellationToken);
     }
 
+    /// <inheritdoc/>
+    public virtual IAsyncEnumerable<T> AsAsyncEnumerable(ISpecification<T> specification)
+    {
+      return dbContext.Set<T>().AsAsyncEnumerable();
+    }
+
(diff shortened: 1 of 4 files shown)


ardalis (approved):

ardalis:
@fiseni any concerns?

fiseni:
I've been hesitant to add the `AsAsyncEnumerable`. The specifications may contain `PostProcessingAction`, and that won't work in this case, since you need the data in memory. I don't know if anyone is using that feature, but if anyone does, then they might get inconsistent behavior.

fiseni:
Btw, I have nothing against this feature being merged. But, if someone complains, we should forward them here :)

SrushtiPasari94:
If I update the Ardalis.Specification to version 7.0.0 form version 6.1.0, I am getting **Method 'AsAsyncEnumerable' in type '**.**.**.**.EfRepository`1' from assembly 'Infrastructure, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.'**

fiseni:
Hi @nkz-soft,

I assume you updated only the `Ardalis.Specification` package. You should update also the `Ardalis.Specification.EntityFrameworkCore` package. The `IRepository` interface has additional method, and the implementations are in the plugin package.

fiseni:
I tagged @nkz-soft by mistake. The message was intended to you @SrushtiPasari94.
You have to update the version for the plugin package.

I saw you were working on EShop project. I commented in your PR there too.

SrushtiPasari94:
@fiseni Updates Ardalis.Specification.EntityFrameworkCore packages as well. Its working as expected. 
Thankyou!

Issue #294, opened 2022-09-22
SelectMany option for Specification<T, TResult>


MGramolini:
It would be useful to be able to do a SelectMany and I believe it would not take much effort to add.  I would be willing to make the adjustments if it would be considered.

MisinformedDNA:
I'm hitting a major performance issue. I'm currently querying an aggregate and its entities, but the DbContext is getting bloated and causing slow performance. My thought was to write two specs, one that queries the aggregate with `AsNoTrackingWithIdentityResolution()` and the other to query the entity, through the aggregate and the use of `SelectMany()`. I can't query the entity directly because my `IRepository` only works with aggregates.

I could ditch `Specification` for this scenario but having `SelectMany()` would make my life much easier.

@ardalis @fiseni ?

fiseni:
Hi,

Yea, it would be great. Wanna give it a shot? @MGramolini ?

It's a bit of a problem though. Users might think that they can chain multiple `SelectMany` and `Select` statements, and that won't be the case. Yes, we're chaining various other methods, but we're just building the specification state. In the case of projections (the `Select` variants), the return type is a new type. But, the input to the next method will still be the entity in our case. So, it will be a bit confusing.

ardalis:
If someone wants to take on adding SelectMany support to the library, we're open to a pull request for it!

amdavie:
I also have a need for SelectMany. I haven't seen any traffic on this for a while, so I'm willing to attempt a PR if no one else has started work on this.

stefannikolei:
Any plans for a release in the near future with thsi feature?

fiseni:
Yes, in the next two weeks, around 15th May.

fiseni:
Closing the issue, the feature is published in version 7.

PR #320, opened 2023-03-31
Implement SelectMany support

amdavie:
For issue https://github.com/ardalis/Specification/issues/294

Usage:
```c#
public class StoreProductNamesSpec : Specification<Store, string?>
{
  public StoreProductNamesSpec()
  {
    Query.SelectMany(s => s.Products.Select(p => p.Name));
  }
}
```

Concurrent usage of Select and SelectMany is prevented via a runtime evaluator exception.

We may want to rename `ISpecification<T, TResult>.Selector` to `SelectExpression` to match the newly added `SelectManyExpression` to be consistent (I couldn't come up with a good name other than SelectManyExpression).

Let me know how it looks - happy to make additional changes.


amdavie:
@fiseni no problem. `SelectorMany` it is.

fiseni (approved):

PR #328, opened 2023-05-15
Add support for extending default evaluator list

### LINKED ISSUE #308: Add support for extending default evaluator list
There exists no possibility to extend the default evaluators listed in `SpecificationEvaluator.Default`. There exists an example of how to write your own evaluator in in https://github.com/ardalis/Specification/issues/247 but it doesn't touch the subject of how to inject it.

If I want to extend with a custom evaluator I've to create my own list of all the default evaluators and then add my custom one. This gives a hard coded dependency and if Ardalis makes a change to the default list of evaluators I will have to update my own code as well to keep up.

Suggestion: at least change the `private readonly List<IEvaluator> evaluators` to `protected` so projects can create their own custom SpecificationEvaluator extending from Ardalis SpecificationEvaluator and add their on evaluators to the default ones.

This would result in:
```csharp
  public class CustomSpecificationEvaluator : SpecificationEvaluator
  {
    public CustomSpecificationEvaluator(bool cacheEnabled = false) : base(cacheEnabled)
    {
      evaluators.Add(AsSingleQueryEvaluator.Instance);
    }
  }
```

instead as of today being forced to code:
```csharp
  public class CustomSpecificationEvaluator : SpecificationEvaluator
  {
    public CustomSpecificationEvaluator(bool cacheEnabled = false) : base(GetAllEvaluators(cacheEnabled))
    {
    }

    private static IEnumerable<IEvaluator> GetAllEvaluators(bool cacheEnabled = false)
    {
      var defaultAndCustomEvaluators = new List<IEvaluator>();
      // add Ardalis default evaluators
      defaultAndCustomEvaluators.AddRange(new IEvaluator[]
      {
                // this list will need to be maintained
                WhereEvaluator.Instance,
                SearchEvaluator.Instance,
                cacheEnabled ? IncludeEvaluator.Cached : IncludeEvaluator.Default,
                OrderEvaluator.Instance,
                PaginationEvaluator.Instance,
                AsNoTrackingEvaluator.Instance,
                IgnoreQueryFiltersEvaluator.Instance,
#if !NETSTANDARD2_0
                AsSplitQueryEvaluator.Instance,
                AsNoTrackingWithIdentityResolutionEvaluator.Instance
#endif
      });

      // add custom evaluators
      defaultAndCustomEvaluators.AddRange(new IEvaluator[]
      {
        AsSingleQueryEvaluator.Instance
      });

      return defaultAndCustomEvaluators;
    }
  }
```

There exists many solutions to this problem. Another one would be to expose the `evaluators` property. Whatever solution that don't force you to copy-paste the default evaluator list is fine. 🙂

fiseni:
closes #308 

The PR enables us to more easily extend the evaluators list. We have exposed the state as protected.

```csharp
public class MyPartialEvaluator : IEvaluator
{
  private MyPartialEvaluator () { }
  public static MyPartialEvaluator Instance { get; } = new MyPartialEvaluator();

  public bool IsCriteriaEvaluator { get; } = true;

  public IQueryable<T> GetQuery<T>(IQueryable<T> query, ISpecification<T> specification) where T : class
  {
    // Write your desired implementation

    return query;
  }
}

public class MySpecificationEvaluator : SpecificationEvaluator
{
  public static MySpecificationEvaluator Instance { get; } = new MySpecificationEvaluator();

  private MySpecificationEvaluator() : base()
  {
    Evaluators.Add(MyPartialEvaluator .Instance);
  }
}
```

### DIFF
--- a/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/Evaluators/SpecificationEvaluator.cs
+++ b/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/Evaluators/SpecificationEvaluator.cs
@@ -10,11 +10,11 @@ namespace Ardalis.Specification.EntityFramework6
     // Will use singleton for default configuration. Yet, it can be instantiated if necessary, with default or provided evaluators.
     public static SpecificationEvaluator Default { get; } = new SpecificationEvaluator();
 
-    private readonly List<IEvaluator> evaluators = new List<IEvaluator>();
+    protected List<IEvaluator> Evaluators { get; } = new List<IEvaluator>();
 
     public SpecificationEvaluator()
     {
-      this.evaluators.AddRange(new IEvaluator[]
+      this.Evaluators.AddRange(new IEvaluator[]
       {
                 WhereEvaluator.Instance,
                 SearchEvaluator.Instance,
@@ -26,7 +26,7 @@ namespace Ardalis.Specification.EntityFramework6
     }
     public SpecificationEvaluator(IEnumerable<IEvaluator> evaluators)
     {
-      this.evaluators.AddRange(evaluators);
+      this.Evaluators.AddRange(evaluators);
     }
 
     /// <inheritdoc/>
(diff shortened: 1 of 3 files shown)


ardalis (approved):

ardalis:
In your example here, how is `MySpecificationEvaluator` discovered? Is there some registration logic or some reflection-based discovery or what?

PR #350, opened 2023-07-23
Create new CI and Release workflows.

### LINKED ISSUE #349: Update pipelines.
The workflows are not particularly nice. The publish workflows are triggered on each push, and they fail since the version is not updated. Also, they're more complicated than necessary.
- We have too many workflows, and this is because we need one for EF6 that runs on a Windows host. If we're having Windows host anyway, then why not run everything on it?
- Running tests on docker is necessary since we need `MSSQL` db for integration tests. But, if we're using a Windows host we don't need that. The images come with `LocalDb` preinstalled.
- The test reports are also overkill. It was really cool having them in Azure pipelines since we could see the reports in a separate tab. But, we don't have that in GitHub Action and we're compiling them for no reason.
- The task we have for publishing packages is outdated. Now we have the `dotnet nuget push` which is way more simpler.

We might have a single workflow `ci.yml` for build/test and triggered on pr and push on main. And an additional one `release.yml` which will be triggered on `Release`, whenever we publish a release on GitHub.

fiseni:
Closes #349. Please check the related issue for more details.
Removed the old workflows and created two new ones. I think the workflows now are stupid simple :)

- `ci.yml` - Build and test, triggered on pr and push on main. Triggered only if there are changes in the library projects. All other changes for docs or sample apps won't trigger anything. It's tedious if CI is triggered on each doc change.
- `release.yml` - Triggered on release publish. We're anyway always creating releases from Github, so let's use that event. This will publish all three packages (but that's what we're always doing). If in the future we need to publish them independently, we might use different tags for each package, but for now, that seems unnecessary.

The solution contains sample apps too, but I didn't want to build them as part of the workflows. So, instead of passing all the required projects as an argument to `dotnet build`, I created a solution filter `ci.slnf` which contains only the necessary projects. I think this was a great trick, and makes the yml files way cleaner.


ardalis:
Sounds great; let's see them in action!

ardalis:
Also, nice use of a solution filter; I probably would have created a separate ci.sln file to do the same thing, but your approach is cleaner I think.

PR #352, opened 2023-07-24
Dummy PR to test the triggers.

fiseni:


fiseni:
@ardalis The PR triggered the correct workflow.

On merge, this will run again. If you protect the `main` branch, so we're sure nobody can push directly to the main, then perhaps we don't need the `push` trigger. It doesn't hurt, but it's just wasting resources.

fiseni:
Actually, we should leave it as it is. We should run it on merge too, so the builds are visible on the commit history. That affects the badges I assume, so better to have it.

PR #354, opened 2023-07-25
Add more samples.

### LINKED ISSUE #344: Update Sample projects
The sample projects are really old and outdated. We should delete them, and write new ones from scratch. We may have multiple projects
- SampleProject1 - The basic usage, utilizing the built-in repositories. Also, show some samples of how to use specs without repos.
- SampleProject2 - How to extend the specifications, builder, and how to put all pieces together.
- SampleProject3 - I want to demonstrate how to combine specs with custom repos, MediatR, Automapper (or any other mapper), etc.

We shouldn't push toward a given architectural style (e.g. clean architecture), that's up to the users. The projects should be as simple as possible, perhaps web projects using minimal API.

fiseni:
Closes #344 
Added more samples and improved the sample applications. Added various usages (single result spec, selectors, extensions, evaluators, paginated results, projections with Automapper, etc.). Added a lot of comments. 

I hope users will find this helpful.


ardalis:
Awesome! I'll try to update the docs to deep link into the samples when I get a chance, too!

PR #357, opened 2023-08-04
Add script for SQL Local DB setup.

fiseni:
The `Local DB` comes pre-installed for Windows hosts, but our tests are occasionally failing to establish a connection. Adding an explicit task might ensure the DB is up and ready before it continues with the next job. At least I think so :) Let's try it.


fiseni:
It seems it fixed the issue.
Btw, we'll have to update and improve the fixtures for our infrastructure tests, but I'll do that in a separate PR.

PR #360, opened 2023-08-05
Refactor and improve integration test fixtures.

### LINKED ISSUE #359: Refactor and improve integration test fixtures.
The fixtures for integration tests are not particularly nice. We created them a long time ago, and they need an update and improvement. Also, there are a few issues in the current setup.
- We're working with a single database. That's not an extendable solution.
- We're using the same `DbContext` instance within a test class. This can lead to unreliable tests, since the change tracker may already contain objects from other test runs.
- For seeding, we're using `HasData` feature of `EF Core`. I'm not particularly a fan of it, since can lead to unreliable tests too.
- The fixtures for `EF Core` and `EF6` are not consistent.

fiseni:
Closes #359. More details on the related issue.

This PR contains the following improvements.
- Created a brand new `DatabaseFixture`. We'll create a unique database per each instance, and we'll delete the database on disposal.
- We seed the database with test data (we actually write data, and not using `HasData`).
- The tests won't reuse a shared `DbContext`, instead, we create a separate context for each test. This will lead to more reliable tests (especially while testing `Include` features).
- We can define collection fixtures now. All tests part of the same collection will work against the same database. For now, we have a single `ReadCollection` that is utilized for all repository read actions.
- For write actions, either we define a separate collection or utilize the `ClassFixture` instead.
- The fixtures for `EF Core` and `EF6` are consistent now.
- Other improvements and cleanup.


fiseni:
Note to self: The connection to the database fails. It happens randomly for a given project, usually 2 of them succeed completely and 1 one of them fails (and they shift randomly). I suspect now that we're creating DbContext for each test separately, we face connection pool exhaustion perhaps?

Btw, locally there is no issue at all; neither running from Test Explorer, nor through cli, nor through Docker. It has to do something with the environment in CI.

```
System.Data.SqlClient.SqlException : Connection Timeout Expired.  The timeout period elapsed while attempting to consume the pre-login handshake acknowledgement.  This could be because the pre-login handshake failed or the server was unable to respond back in time.  The duration spent while attempting to connect to this server was - [Pre-Login] initialization=20164; handshake=24; 
```

PR #369, opened 2023-09-02
Talpers/delete range by spec

thorstenalpers:
New feature DeleteRange by specification

### DIFF
--- a/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
+++ b/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
@@ -81,6 +81,15 @@ public abstract class RepositoryBase<T> : IRepositoryBase<T> where T : class
     }
 
     /// <inheritdoc/>
+    public virtual async Task DeleteRangeAsync(ISpecification<T> specification, CancellationToken cancellationToken = default)
+    {
+        var query = ApplySpecification(specification);
+        _dbContext.Set<T>().RemoveRange(query);
+
+        await SaveChangesAsync(cancellationToken);
+    }
+
+    /// <inheritdoc/>
     public virtual async Task<int> SaveChangesAsync(CancellationToken cancellationToken = default)
     {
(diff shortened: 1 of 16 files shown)


fiseni:
Hi @thorstenalpers,

Thank you for your work. It looks great.
There is a slight issue in the integration tests (e.g. `RepositoryOfT_DeleteRangeAsync`). You correctly created a new collection `WriteCollection`, but all methods within that class will run against the same fixture and same database. So, if the first method removed that entity successfully, the rest of the methods actually won't do anything (there is no such record anyway), hence we have a false sense of security.

I'd suggest, deleting a different entity in each test (passing a different id). I think we're seeding many products, so just use 1,2,3, and so on.

Let me know if you face any issues.

thorstenalpers:
Hi @fiseni, thanks for the help. Yes, I have an issue with EF6 and deleting entities using specs that use .AsNoTracking() to query entities. With EfCore the test with deletion was successful.

Failed Test: DeletesProductWithIdOne_GivenProductByIdAsUntrackedSpec()
```
 await repo.DeleteRangeAsync(new ProductByIdAsUntrackedSpec(4));

 //   System.InvalidOperationException : The object cannot be deleted because it was not found in the ObjectStateManager.
```

```
public class ProductByIdAsUntrackedSpec : Specification<Product>, ISingleResultSpecification
{
    public ProductByIdAsUntrackedSpec(int id)
    {
        Query.Where(product => product.Id == id).AsNoTracking();
    }
}
```
For me a deletion by a query as untracked makes no sense either. Should I remove these tests?

fiseni:
Hey @thorstenalpers,

It looks fine, thanks. It's a little bit dubious that we're not explicitly loading the entities (e.g. use ToListAsync, or LoadAsync), but we're passing an `IQueryable` to `RemoveRange`. So, we're relying completely on the internal mechanism of EF to load the entities (which may change). But, I think it's ok. Let's merge this.

```csharp
var query = ApplySpecification(specification);
_dbContext.Set<T>().RemoveRange(query);
```

fiseni (approved):

hgedik:
Could you please update Nuget as well? :)

PR #391, opened 2024-03-16
Fix InMemory SearchExtension bug

### LINKED ISSUE #390: InMemory SearchExtension bug
There is a bug in the `SearchExtension` class used by the InMemory SearchEvaluator.

When the input string contains the requested pattern, BUT the first character of said pattern is used in the power of 2, the result is false.

Note the amount of 'a' characters in the screenshot

image

fiseni:
Fixes #390


ardalis (approved):

ardalis:
That was a weird bug... Thanks for fixing!

PR #398, opened 2024-05-09
397 make fields and methods protected

### LINKED ISSUE #397: Make repository methods protected to allow for subclassing
In order to customise and extend the behaviour of the Repository (RepositoryBase), its methods and fields should be protected rather than private

eldamir:
Fixes #397

### DIFF
--- a/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
+++ b/Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
@@ -10,25 +10,25 @@ namespace Ardalis.Specification.EntityFramework6;
 /// <inheritdoc/>
 public abstract class RepositoryBase<T> : IRepositoryBase<T> where T : class
 {
-    protected readonly DbContext _dbContext;
-    protected readonly ISpecificationEvaluator _specificationEvaluator;
+    protected DbContext DbContext { get; private set; }
+    protected ISpecificationEvaluator SpecificationEvaluator { get; private set; }
 
     public RepositoryBase(DbContext dbContext)
-        : this(dbContext, SpecificationEvaluator.Default)
+        : this(dbContext, EntityFramework6.SpecificationEvaluator.Default)
     {
     }
 
     /// <inheritdoc/>
     public RepositoryBase(DbContext dbContext, ISpecificationEvaluator specificationEvaluator)
     {
-        _dbContext = dbContext;
-        _specificationEvaluator = specificationEvaluator;
+        DbContext = dbContext;
+        SpecificationEvaluator = specificationEvaluator;
     }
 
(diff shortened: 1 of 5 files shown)


Specification.EntityFramework6/src/Ardalis.Specification.EntityFramework6/RepositoryBaseOfT.cs
```diff
@@ -10,8 +10,8 @@ namespace Ardalis.Specification.EntityFramework6;
 /// <inheritdoc/>
 public abstract class RepositoryBase<T> : IRepositoryBase<T> where T : class
 {
-    private readonly DbContext _dbContext;
-    private readonly ISpecificationEvaluator _specificationEvaluator;
+    protected readonly DbContext _dbContext;
```
enrij, line 13: I think a property

`protected DbContext { get; private set; }`

combined with 

```
public RepositoryBase(DbContext dbContext)
{
    DbContext = dbContext;
}
```

is more appropriate then a protected member.

The same for `SpecificationEvaluator`

eldamir: Didn't want to change too much, but I see your point. If we let subclasses access these "fields" it will be safer to do "props" instead from a maintenance perspective. I've made that change and pushed

ardalis:
If you use a property with `private set;` doesn't that still limit extensibility since there is no way to replace a dependency with a different implementation? Sure, readonly access is better than no access, but I thought the intent of #397 was to allow the replacement of the dependency if desired...

eldamir:
For my needs, I only needed to override the methods while still being able to refer to and read the fields (now props) from the subclass. I don't need to mess with the constructor signatures or replace/modify the state. Only the behaviour.

So for me, this is enough. I can see a case for opening it up further. If you believe this is the way to go, I'll happily remove the `private` markers 😊

ardalis:
I'm not clear on what the benefit of the property are over the field, in this case. The field change (as originally done in this PR) seems simpler and provides greater extensibility. The property adds additional encapsulation, but I'm not sure that's needed or desirable in this case.

However, I recognize I'm not the target audience for this change, so if y'all are ok with it as is I think we can merge it. Let me know.

eldamir:
In that case, let's opt for maximum extensibility with maximum encapsulation. Most power and most safety. I've removed the `private` keyword to make the properties `protected` on both get and set

enrij (approved):
See comment [here](https://github.com/ardalis/Specification/pull/398#issuecomment-2109348077)

eldamir:
I think we landed in the best spot here. No "field inheritance", and properties are available for read and write as needed by inheritors with proper encapsulation. I'd be happy to see this merged 😉

ardalis:
Now I just need to release it...

gocampo:
Just what I needed!! Waiting the release

PR #402, opened 2024-08-04
Primary constructor implement

KhanbalaRashidov:


fiseni:
Thanks. But, we would use primary constructors sparingly, whenever it makes sense.
We're already in the process of refactoring the solution. I'll close this PR, sorry.

PR #440, opened 2025-02-24
Update GitHub workflows to use Linux hosts.

fiseni:
We've been using Windows hosts and LocalDb database for the CI workflows. That's mainly because of the EF6 package and `net472` target.

But it's not stable, and we have random fails. The connection to localDb fails because it's not always up and ready. We can mitigate the issue by waiting for the database to be ready before running the tests.

Anyhow instead of all that, since the EF6 package targets `net9.0` too; then let's just use ubuntu hosts and run tests using only `net9.0` runtime. We're using TestContainers, so we won't need localDb, we'll utilize the mssql container for the tests.

It's a bit of a risk not to run the tests for `net472` target, but it's a rarely changing package. And also we may always run it locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   build:
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     
     steps:
     - name: Checkout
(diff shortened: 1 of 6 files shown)


ardalis:
Agreed.

ardalis (approved):

Issue #418, opened 2024-11-09
Remove the state for the Query builder. Update it as a calculated property.


fiseni:
We no longer will keep the specification builder a state. It will be defined as a calculated property and we'll create an instance on access `public ISpecificationBuilder<T> Query => new SpecificationBuilder<T>(this);`

In version 9 we'll improve the builders, so the whole query can be written as a single chain. Therefore the common use will not require more than one allocation. By removing the state, we'll save an additional 8 bytes used for the reference.

Other than that, the property will no longer be defined as virtual. Not a single user is overriding and defining their specification builder. The builder is the essence of this library.

Issue #417, opened 2024-11-09
Remove the Evaluator and Validator state from Specifications.


fiseni:
The specifications were designed to be flexible and extendable. The users may choose to replace evaluators, write their extensions, and so on. In reality, 95% of the users do not do that, and they should not pay the price of this "extensibility".

The theme of version 9 is to reduce the allocation as much as possible. In that context, the specifications no longer accept custom implementations of `IInMemorySpecificationEvaluator` and `ISpecificationValidator` as constructor arguments. Instead of keeping these implementation references as a state (and wasting 16 bytes), they will be defined as calculated properties. Our default implementations are singleton instances anyway, and we won't have additional allocations on access. Anyhow, the properties will be defined as virtual, and users will still be able to alter the behavior and use their implementations if necessary.

**Breaking changes:**
The common use of this library will remain unchanged. Folks who use custom in-memory evaluators or validators will need to change the way they wire them up.

Issue #416, opened 2024-11-09
Refactor Take/Skip state to non-nullable int. Define the default value to -1.


fiseni:
The theme of version 9 is to reduce the allocation as much as possible. We'll squeeze every unnecessary allocated byte.
In this context, `Take` and `Skip` properties are defined as nullable int values, where the null value means "not set". Considering that `int?` has a size of 8 bytes compared to 4 bytes for `int` (on x64), we'll update these properties to non-nullable int, and the default value will be defined as -1.

**Breaking change:**
The common use of this library will remain unchanged. But, folks who have custom evaluators, extensions, or have custom logic that depends on the specification state directly will be affected by this change.

PR #441, opened 2025-02-26
Reduce the size of specifications and avoid unnecessary memory allocations

### LINKED ISSUE #415: Reduce the Specification size. Set the initial state to null.
The specification state is not optimized at all and we have a lot of unnecessary memory allocations. This is partly because many users have assumptions about the internals and we have avoided breaking user code (even though technically it's not a breaking change). But, it's time to change that. For the upcoming version 9, we'll be not so aggressive, and we won't refactor everything from the ground up. We'll try to keep the public API mostly the same, but it will be a breaking change for the folks who've been relying on the internals.

The specification state mostly consists of various expression collections. Currently, all these collections are initialized during instantiation. So, we have several list objects regardless if they're used by the spec or not. This is wasteful, a single empty list allocates 56 bytes of memory.

In version 9 we'll set all the initial states to null. We'll still be spending 8 bytes per reference (on x64), it's not ideal and we have further plans for version 10. But, taking this incremental step would reduce the size of specifications considerably. The aim is to keep the size ~100 bytes for empty specs (currently is over 392 bytes). Also, users shouldn't pay for features they don't use.

**Breaking Changes:**
The public behavior will remain unchanged. The exposed `IEnumerable` properties will return Empty collections.
However, users who depend on the internals and assume the underlying collection is `List` and is initialized, would need to update their code. For instance, some users have taken shortcuts and do this `((List<WhereExpressionInfo<T>>)spec.WhereExpressions).Add(...)`. This would be a breaking change for them.

fiseni:
Fixes #415 #416 #417 #418 

TLDR: The size of an empty is reduced to 96 bytes, from 392 bytes.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,6 +13,7 @@
   </ItemGroup>
   
   <ItemGroup Label="Test projects dependencies">
+    <PackageVersion Include="ManagedObjectSize.ObjectPool" Version="0.0.7-gd53ba9da59" />
     <PackageVersion Include="MartinCostello.SqlLocalDb" Version="3.4.0" />
     <PackageVersion Include="Microsoft.EntityFrameworkCore.SqlServer" Version="9.0.2" />
     <PackageVersion Include="Respawn" Version="6.2.1" />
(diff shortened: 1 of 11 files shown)


ardalis (approved):

ardalis:
:shipit:

PR #442, opened 2025-02-27
Update and improve the builder infrastructure.

### LINKED ISSUE #419: Improve the specification builders. Select/SelectMany should be applied in the end.
The current builder infrastructure has an annoying flaw. The `Select`/`SelectMany` expressions can be applied only at the start of the query. This forces users to have an additional `Query` just for the select expressions. It doesn't feel natural and creates a lot of confusion.

We should refactor the infrastructure so the select expressions can be applied at the end of the query. That's more idiomatic, mimics LINQ better, and it's simply the correct way. Implementation-wise, this will require us to have two completely separate chains, one for `ISpecificationBuilder<T>` and another for `ISpecificationBuilder<T, TResult>`. This implies that we have to duplicate all extension methods. It's a bit tedious work and harder to maintain, but it's worth it.

Other than that, the select extensions should break the chain (they will return void), and no further chaining will be allowed on top of them. This more accurately describes our builders and avoids the assumption that any further expressions are applied to the projected result (which is not the case, we're not operating on IEnumerable/IQueryable).

**Breaking Changes:**
- The `Select`/`SelectMany` are applied at the end of the chain or in a separate `Query` clause. These extension methods return void and no further chaining is possible.
- Folks who have custom extensions, should align to the new infrastructure and create extensions methods for both `ISpecificationBuilder<T>` and `ISpecificationBuilder<T, TResult>`

### LINKED ISSUE #420: Add IsChainDiscarded thread-static field.
All builder extension methods have an overload that accepts a `bool condition` parameter. If the passed condition is `false` the expression/value is not added to the specification state. This is convenient in many scenarios and is similar to `WhereIf` types of LINQ extensions.

However, having this condition complicates the case where we have sub-chains (e.g. Include/ThenInclude, OrderBy/ThenBy, etc). If any parent in the sub-chain has a false condition, we need to dismiss the rest of the sub-chain regardless of their conditions. To implement this we had to define the `OrderedSpecificationBuilder`, `IncludableSpecificationBuilder` and `CacheSpecificationBuilder` sub-classes to `SpecificationBuilder`, with the sole purpose of "remembering" the previous condition of the sub-chain. The root parent of the chain returns new instances of these builders which act as a container for the sub-chain. This is a bit wasteful, we're spending 32 bytes per instance.

Considering that the constructor will be executed in a single thread (and generally specifications are not intended to be thread-safe), then we may introduce a ThreadStatic field in the specification itself that indicates whether the chain is discarded or not. It will be utilized only during the building stage for the sub-chains. Once the state is built, we don't care about it anymore. The initial value is not important since the value is always initialized by the root of the sub-chain. Therefore, we don't need ThreadLocal (it's more expensive). 

We still need the `IncludableSpecificationBuilder` to remember the generic parameter of the parent, but we won't need the `OrderedSpecificationBuilder` and `CacheSpecificationBuilder` instances. This will save us 8 bytes per include builder, and 32 bytes per order/cache builder instances.

**Breaking changes:**
- No breaking changes for common usage of the library.
- Folks that have custom extensions to Include/Order/Cache should adapt their implementations.

fiseni:
Fixes #419 
Fixes #420


src/Ardalis.Specification/Builders/Builder_Flags.cs
```diff
@@ -0,0 +1,328 @@
...
+    /// </summary>
+    /// <typeparam name="T">The type of the entity.</typeparam>
+    /// <param name="builder">The specification builder.</param>
+    /// <param name="condition">The condition to evaluate.</param>
+    /// <returns>The updated specification builder.</returns>
+    public static ISpecificationBuilder<T> IgnoreQueryFilters<T>(
+        this ISpecificationBuilder<T> builder,
+        bool condition) where T : class
+    {
+        if (condition)
```
ardalis, line 57: What's the purpose of passing in a condition flag?

ardalis (approved):

src/Ardalis.Specification/Builders/Builder_Flags.cs
fiseni: It's similar to WhereIf types of LINQ extensions. In our case, if the condition is false, the statement does nothing, meaning the expression/value is not added to the specification state. This allows you to have a more succinct way of expressing the query. Instead of having if clauses, you can pass that condition to the extension method itself. We have this for a long time.
The #420 contains more details

PR #443, opened 2025-03-01
Improve search validator and in-memory evaluator.

### LINKED ISSUE #421: Improve the C# implementation of the Like operator.
A detailed description will be added soon.

### LINKED ISSUE #422: Improve the Search in-memory evaluator. Implement a custom iterator and reduce allocations.
The current `SearchMemoryEvaluator` implementation is as follows.

```csharp
public IEnumerable<T> Evaluate<T>(IEnumerable<T> query, ISpecification<T> specification)
{
    foreach (var searchGroup in specification.SearchCriterias.GroupBy(x => x.SearchGroup))
    {
      query = query.Where(x => searchGroup.Any(c => c.SelectorFunc(x).Like(c.SearchTerm)));
    }

    return query;
}
```

This looks relatively simple, but there are a lot of hidden allocations. Moreover, the number of allocations is not constant, and it depends on the size of the input collection.

We need to implement a custom iterator and come up with an allocation-free GroupBy implementation. We may keep the search expressions sorted by search group, and then slice to get the groups.

**Breaking Changes:**
The `SearchEvaluator` was renamed to `SearchMemoryEvaluator`. It was causing unnecessary name collisions with EF plugin evaluators.

### LINKED ISSUE #424: Refactor Search validator to allocation-free implementation.
The current `SearchValidator` implementation is as follows.

```csharp
public bool IsValid<T>(T entity, ISpecification<T> specification)
{
    foreach (var searchGroup in specification.SearchCriterias.GroupBy(x => x.SearchGroup))
    {
        if (searchGroup.Any(c => c.SelectorFunc(entity).Like(c.SearchTerm)) == false) return false;
    }

    return true;
}
```

This looks relatively simple, but there are a lot of hidden allocations. Moreover, the number of allocations is not constant, and it depends on the size of the input collection.

We need allocation-free GroupBy implementation. We may keep the search expressions sorted by search group, and then slice to get the groups.

fiseni:
Fixes #421 
Fixes #422 
Fixes #424 

- Improved the nullability in search expressions.
- Improved the C# implementation of Like. Introduced caching for regex objects.
- Updated `SearchValidator` to allocation-free implementation. For larger input collections, the difference is drastic. For 1000 entities, the allocations are reduced from **651.160 bytes to ZERO bytes.**
- Updated `SeatchMemoryEvaluator` and decreased allocations (constant allocation size regardless of input). Added custom iterator implementation. For larger input collections, the difference is drastic. For a source of 1000 items, the allocations are reduced from **257.872 bytes to only 56 bytes** (the cost of the custom iterator instance).


src/Ardalis.Specification/Evaluators/SearchExtension.cs
```diff
@@ -1,42 +1,81 @@
...
+        return regex;
+    }
+
     public static bool Like(this string input, string pattern)
     {
         try
         {
-            return SqlLike(input, pattern);
+            // The pattern is dynamic and arbitrary, the consumer might even compose it by an end-user input.
+            // We can not cache all Regex objects, but at least we can try to reuse the most "recent" ones. We'll cache 10 of them.
```
ardalis, line 32: Makes sense

src/Ardalis.Specification/Evaluators/SearchMemoryEvaluator.cs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private sealed class SpecLikeIterator<TSource> : Iterator<TSource>
+    {
+        private readonly IEnumerable<TSource> _source;
+        private readonly List<SearchExpressionInfo<TSource>> _searchExpressions;
+
+        private IEnumerator<TSource>? _enumerator;
+
+        public SpecLikeIterator(IEnumerable<TSource> source, List<SearchExpressionInfo<TSource>> searchExpressions)
```
ardalis, line 28: Not switching to primary constructor syntax yet?

src/Ardalis.Specification/Specification.cs
```diff
@@ -84,10 +84,30 @@ public class Specification<T> : ISpecification<T>
...
     internal void Add(WhereExpressionInfo<T> whereExpression) => (_whereExpressions ??= new(2)).Add(whereExpression);
-    internal void Add(SearchExpressionInfo<T> searchExpression) => (_searchExpressions ??= new(2)).Add(searchExpression);
     internal void Add(OrderExpressionInfo<T> orderExpression) => (_orderExpressions ??= new(2)).Add(orderExpression);
     internal void Add(IncludeExpressionInfo includeExpression) => (_includeExpressions ??= new(2)).Add(includeExpression);
     internal void Add(string includeString) => (_includeStrings ??= new(1)).Add(includeString);
+    internal void Add(SearchExpressionInfo<T> searchExpression)
+    {
+        if (_searchExpressions is null)
+        {
+            _searchExpressions = new(2) { searchExpression };
```
ardalis, line 94: Replace magic number `2` with meaningful named constant?

ardalis (commented):
Looks good; couple minor questions

src/Ardalis.Specification/Specification.cs
fiseni: This represents the initial capacity. And it can be different for each collection, for example we may choose for Where to be 4, for Search 2, etc. We need to introduce 6-7 constants, and not sure whether that makes it better or worse.

src/Ardalis.Specification/Evaluators/SearchMemoryEvaluator.cs
fiseni: I've been hesitant because of the older TFMs. We need to check whether we need any polyfills or not (I assume not).

src/Ardalis.Specification/Specification.cs
fiseni: Updated to use constants.

PR #454, opened 2025-03-11
Add tests

fiseni:
- Full 100% coverage for the base package (line and branch).
- Full coverage for the EF Core package (line and branch), except `ContextFactoryRepositoryBaseOfT`. I don't think we should even have this one.


ardalis (approved):

ardalis:
Nice! Yah they order issue should wait but it would be good if we had a way to opt into a fix sooner. Or just ship v10 a week after 9 and then start work on 11...

PR #472, opened 2025-05-26
Docs - How to Create Specifications

fiseni:


sadukie (changes requested):
Minor issue with a sentence, but otherwise looks good.

PR #549, opened 2026-04-28
Refactor: Syntax improvements

aniekanBane:


fiseni:
This is not just an innocent syntax cleanup. There are bugs. This seems like AI generated content.
I'm closing this PR, sorry.

aniekanBane:
Not AI generated content. Mostly Editor Warnings and Suggestions.

But if you could maybe point out some of the bugs for my own reference that would be nice.

Thanks

src/Ardalis.Specification/Validators/SearchValidator.cs
```diff
@@ -46,28 +46,35 @@ private static bool IsValid<T>(T entity, List<SearchExpressionInfo<T>> list)
...
-                groupStart = i;
+                continue;
+            }
+
+            if (IsValidInOrGroup(entity, list, groupStart, i) is false)
+            {
+                return false;
             }
+
+            groupStart = i;
```
fiseni, line 59: This is not better for me, regardless of what IDEs suggest. Neither simplifies nor decreases the cognitive load.

src/Ardalis.Specification/Validators/SpecificationValidator.cs
```diff
@@ -23,11 +23,6 @@ public SpecificationValidator(IEnumerable<IValidator> validators)
 
     public virtual bool IsValid<T>(T entity, ISpecification<T> specification)
     {
-        foreach (var partialValidator in Validators)
-        {
-            if (partialValidator.IsValid(entity, specification) == false) return false;
-        }
-
-        return true;
+        return Validators.All(partialValidator => partialValidator.IsValid(entity, specification));
```
fiseni, line 26: I don't want to use LINQ here, it captures the `entity` and `specification` in a closure instance.

src/Ardalis.Specification/Specification.cs
```diff
@@ -32,21 +34,22 @@ public class Specification<T> : ISpecification<T>
     // The initial value is not important since the value is always initialized by the root of the chain.
     // Therefore, we don't need ThreadLocal (it's more expensive).
     // With this we're saving 8 bytes per include builder, and we don't need an order builder at all (saving 32 bytes per order builder instance).
-    [ThreadStatic]
-    [EditorBrowsable(EditorBrowsableState.Never)]
+    [ThreadStatic] [EditorBrowsable(EditorBrowsableState.Never)]
```
fiseni, line 37: What's the point of this? Why all attributes in a single line? I like one attribute per line.

src/Ardalis.Specification/Specification.cs
```diff
@@ -32,21 +34,22 @@ public class Specification<T> : ISpecification<T>
...
-    private OneOrMany<OrderExpressionInfo<T>> _orderExpressions = new();
-    private OneOrMany<IncludeExpressionInfo> _includeExpressions = new();
-    private OneOrMany<string> _includeStrings = new();
-    private OneOrMany<string> _queryTags = new();
+    private OneOrMany<WhereExpressionInfo<T>> _whereExpressions;
+    private OneOrMany<SearchExpressionInfo<T>> _searchExpressions;
+    private OneOrMany<OrderExpressionInfo<T>> _orderExpressions;
+    private OneOrMany<IncludeExpressionInfo> _includeExpressions;
+    private OneOrMany<string> _includeStrings;
+    private OneOrMany<string> _queryTags;
```
fiseni, line 47: Why? I like them to be explicitly instantiated.

fiseni (commented):
Some of the casts are intentional, especially I like having them explicit for the older runtimes.

Other than that, there are too many stylistic changes that actually I don't even agree with. There are few that are reasonable, but everything is crumbled into one peace. I posted few examples.

Thank you for the time and contribution, but it doesn't add value to me. I'm sorry.

src/Ardalis.Specification/Specification.cs
aniekanBane: I don't like this too. I noticed that I misspelled the config settings in my `.editiorconfig` afterwards.

